PR #7, opened 2018-03-26
OEUI-45: Get order configurations from REST API resource

flacode:
### JIRA TICKET NAME 
[OEUI-45: Fetch form configuration from API resource](https://issues.openmrs.org/browse/OEUI-45)
## Summary:
Fetch order form configurations from the REST API resource that will be used to populate some form fields required for order creation.
*Note*: I was working on another machine and I did not reconfigure the git configurations. That's why another name is appearing in my commits. But I have resolved that.


.jshintrc
```diff
@@ -1,3 +0,0 @@
-{
-    "esversion": 6
-}
```
larrystone: Please fix this eof issue. Thanks

PR #8, opened 2018-03-26
OEUI-49: Implement Tab functionality

flacode:
### JIRA TICKET NAME 
[OEUI-49: Implement Tab functionality](https://issues.openmrs.org/browse/OEUI-49)
## Summary:
Implement the tab functionality for switching between patient's careSettings that is `outpatient` and `inpatient`.

### DIFF
--- a/app/js/components/orderEntry/SearchAndAddOrder.jsx
+++ b/app/js/components/orderEntry/SearchAndAddOrder.jsx
@@ -2,67 +2,32 @@ import React from 'react';
 import ActiveOrders from './ActiveOrders';
 import PastOrders from './PastOrders';
 import AddForm from './addForm/AddForm';
+import Tabs from '../tabs/Tabs';
+import Tab from '../tabs/Tab';
 
 class SearchAndAddOrder extends React.Component {
+  state = {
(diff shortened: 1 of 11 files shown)


app/js/components/orderEntry/SearchAndAddOrder.jsx
```diff
@@ -63,7 +87,7 @@ class SearchAndAddOrder extends React.Component {
           </div>
           <br />
         </div>
-      </div>
+      </div> */
```
Betty-Kebenei: Why do you have this code commented?

flacode: @Betty-Kebenei, the code is just for demonstrating how to use the tab. So I commented out the bootstrap implementation. I want to first seek the team's approval before deleting.

geofrocker: I think we no longer need the bootstrap implementation of the tabs and so it can safely removed.

Betty-Kebenei: I also think we can remove it.

PR #9, opened 2018-03-28
OEUI-32: Consume already existing Api to search for a drug

larrystone:
## **JIRA TICKET NAME:**
[OEUI-32: Consume already existing Api to search for a drug](https://issues.openmrs.org/browse/OEUI-32)

### SUMMARY:
This should include real-time autogenerate search for existing drugs. This would also include the front end implementation of the UI using React.


app/js/reducers/drugSearchReducer.js
```diff
@@ -0,0 +1,36 @@
+import * as actionTypes from '../actions/actionTypes';
+import mockData from '../../../__mocks__/mockData';
```
Betty-Kebenei: Why do you need mockData in this reducer?

geofrocker: @Betty-Kebenei We are using mock data as the initial state

larrystone: @geofrocker thanks for assisting with the explanation, 
@flacode yes, mockData was used system wide for providing mock data for usage in both tests(actions/components) and initial values for reducers

app/js/actions/drug.js
```diff
@@ -0,0 +1,37 @@
...
+    .then((response) => {
+      dispatch(searchDrugLoading(false));
+      dispatch(searchDrugSuccess(response.data.results));
+    })
+    .catch((error) => {
+      dispatch(searchDrugLoading(false));
+      if (error.response) {
+        dispatch(searchDrugFailure(error.response));
+      } else {
+        dispatch(searchDrugFailure({ data: { message: "Network connection error" } }));
```
dkayiwa: How do you tell that this is a network connection error?

webpack.sample.config.js
```diff
@@ -43,164 +43,168 @@ let localOwaFolder;
...
-	}
+	var config;
+
+	try {
+		// look for config file
+		config = require('./config.json');
+	} catch (err) {
+		// create file with defaults if not found
+		config = {
+			'LOCAL_OWA_FOLDER': '/Users/name/openmrs-standalone-2.4/appdata/owa/',
```
dkayiwa: Did you intentionally include openmrs-standalone-2.4 in the url?

webpack.sample.config.js
```diff
@@ -43,164 +43,168 @@ let localOwaFolder;
...
+			'APP_ENTRY_POINT': 'http://localhost:8081/openmrs-standalone/owa/openmrs-owa-orderentry/index.html'
```
dkayiwa: Did you intentionally include openmrs-standalone in the url?

larrystone: Okay, this is a sample webpack config that should be used by anyone trying to clone this project. A detailed instruction on how to use this file (and modify) it is given in the readme section of the project. Thanks

webpack.sample.config.js
larrystone: Seen, please refer the next comment below for more information

app/index.html
```diff
@@ -19,8 +19,8 @@
 
   <link rel="stylesheet" href="../../ms/uiframework/resource/appui/styles/header.css" type="text/css"
   />
-  <link rel="stylesheet" href="/openmrs-standalone/ms/uiframework/resource/uicommons/styles/styleguide/jquery-ui-1.9.2.custom.min.css" type="text/css"/>
-
+  <link rel="stylesheet" href="../../ms/uiframework/resource/uicommons/styles/styleguide/jquery-ui-1.9.2.custom.min.css" type="text/css"/>
+  <link rel="stylesheet" href="../../ms/uiframework/resource/orderentryui/styles/drugOrders.css?cache=1521812153298" type="text/css"/>
```
kodero: Is this `cache` query param automatically generated? If it is, all is good. Otherwise, if we are trying to prevent the browser from caching the css resource, then it needs to be dynamically generated.

larrystone: Okay, It wasn't autogenerated, I will remove it.

app/js/actions/drug.js
larrystone: Based on some research on axios, I figured that when there is an error with an axios call and there is no corresponding error.response, it means that the request was not successfully made to the server -  a situation caused by network connection failure

larrystone:
@dkayiwa, yes, all the changes made it possible to fulfil the requirements defined in the user story for the UI and functional implementation of the search drug feature

kodero:
@dkayiwa, Are there any other concerns with this?

PR #13, opened 2018-04-09
OEUI-56: fix careSetting undefined error

fuchodeveloper:
[OEUI-56: Integrate care settings into tab functionality](https://issues.openmrs.org/projects/OEUI/issues/OEUI-56)

SUMMARY
Fix `careSetting` undefined error


app/js/components/orderEntry/OrderEntryPage.jsx
```diff
@@ -1,17 +1,47 @@
...
       </div>
     );
   }
 }
 
-export default OrderEntryPage;
+const mapStateToProps = ({ careSettingReducer }) => ({
+  inpatientCareSetting: careSettingReducer.patientCareSetting,
+  outpatientCareSetting: careSettingReducer.patientCareSetting,
+});
```
zeze111: is there a reason why you separated the patientCareSettings, if they're both receiving the same data from the reducer?

fuchodeveloper: Have modified my implementation to allow for fetching `inpatientCareSetting` and `outpatientCareSetting` explicitly. 
Thanks for pointing that out.

app/js/components/orderEntry/OrderEntryPage.jsx
```diff
@@ -1,17 +1,47 @@
...
+  componentDidMount() {
+    this.props.fetchPatientCareSetting();
+  }
 
-class OrderEntryPage extends React.Component {
   render() {
+    if (!(this.props.outpatientCareSetting || this.props.inpatientCareSetting)) {
+      return (
+        <div>
+          loading...
```
flacode: @fuchodeveloper, I like the way you are using one endpoint for both careSettings, though the `<div>loading</div>` would look better visually say using a spinner than verbal.

fuchodeveloper: Will consult the team concerning the idea of using a reusable spinner in scenarios such as this and update this PR based on the team's response.

fuchodeveloper: Updated now to make use of the default OpenMRS spinner image based on the team's decision.

fuchodeveloper:
@dkayiwa 
Yes. The PR aims to solve the error while also optimizing the code better.

dkayiwa:
I would use separate pull requests for each.

fuchodeveloper:
@dkayiwa 
If I understand correctly, you recommend I split the fix and optimization into two pull requests.

fuchodeveloper:
Ok
I would make the necessary changes

fuchodeveloper:
Closing this PR based on feedback from @dkayiwa. The PR containing the implementations from the feedback can be found [here](https://github.com/openmrs/openmrs-owa-orderentry/pull/21)

PR #14, opened 2018-04-09
Refactor webpack.config.js

mogoodrich:
This pull request renamed "webpack.config.sample.js" to "webpack.config.js" and updated the README.

I don't believe each developers need to manually copy/configure "webpack.config.js".  For overriding/setting "LOCAL_OWA_FOLDER" and "APP_ENTRY_POINT", a developer should just create or customize the "config.js" file (which does not get committed to source control).


mogoodrich:
@dkayiwa  I haven't looked at them yet, but probably should be straightforward to me to resolve... I'm basically just moving webpack.sample.config.js to webpack.config.js.

However, I didn't want to just merge this in because it will likely break people's current development.  I think we should be able to directly commit the “webpack.config.js” file to the project and not force the developer to manually copy “webpack.config.sample.js” to “webpack.config.js”. Customizing the “LOCAL_OWA_FOLDER” and “APP_ENTRY_POINT” configuration parameters can be done by changing/adding a “config.json” file (which would be different for each developer and not committed to source control).

Can I just make this change and merge it in?  Is it enough to post to Talk re: this change... I know there are a lot of people currently developing this module and I didn't want to cause issues. Thoughts?

dkayiwa:
Looks right to commit the change and follow it up with a talk post.

mogoodrich:
Thanks @dkayiwa ... I should be able to follow up tomorrow or Thursday.

mogoodrich:
I got delayed doing this.. :)  Will work on it today but since this is now stale will commit directly to master and let people know...

PR #17, opened 2018-04-10
OEUI-59: Implement accordion UI/UX for active and past orders

larrystone:
## **JIRA TICKET NAME:**
[OEUI-59: Implement accordion UI/UX for active and past orders](https://issues.openmrs.org/browse/OEUI-59)

### SUMMARY:
This includes implementing accordion for both active and past orders in order to reduce the overall height of the web page. With this, a user can click on the active or past drug order tab name to view them and click on the tab name again to toggle/close the expanded view when not needed.

### Example usage:
```
<Accordion 
  open={true} 
  border={false} 
  title="Active Orders"
>
  <h2>Your content here</h2>
</Accordion>
```


app/js/components/accordion/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+      </div>
+    );
+  }
+}
+
+
+Accordion.propTypes = {
+  border: PropTypes.bool,
+  children: PropTypes.shape.isRequired,
```
Betty-Kebenei: Is this supposed to be `shape({})`?

larrystone: Oh yes, it is, will update. Thanks

app/js/components/accordion/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+      </div>
+    );
+  }
+}
+
+
+Accordion.propTypes = {
+  border: PropTypes.bool,
+  children: PropTypes.shape({}).isRequired,
```
segunolalive: When using `PropTypes.shape`, you should actually define the shape of the expected object.
Here's an example from the docs:

```// An object taking on a particular shape
  optionalObjectWithShape: PropTypes.shape({
    color: PropTypes.string,
    fontSize: PropTypes.number
  })

Comment (larrystone) [reply app/js/components/accordion/index.jsx]: Thanks Segun. I will work on this

**Comment by dkayiwa** [comment]
Great point @segunolalive 👍

File: app/js/components/accordion/index.jsx [hunk 3]
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import '../../../css/accordion.css';
```
fuchodeveloper: Could this implementation be added as part of the `openmrs-owa-orderentry.css` which is bundled by webpack so we don't have to be making individual imports?

larrystone: Great suggestion but my concern with that approach is about the conflicts that might arise if/when there are class names that are conflicting classNames in the `openmrs-owa-orderentry.css` and there is also the argument for the concept of code/functional modularity.
However, if the team agrees we implement it this way, then I will gladly accept @Betty-Kebenei @segunolalive @fuchodeveloper @zeze111 @geofrocker @flacode

larrystone: Okay, so the team has agreed that we have all the css styles in one place and a Jira ticket for this purpose has been created because this change has to be implemented in a couple of places/files. Please look forward to that implementation in the future. Thanks

PR #19, opened 2018-04-11
OEUI-61: Standard Dose UI

flacode:
### JIRA TICKET NAME 
[OEUI-61: Standard Dosing UI](https://issues.openmrs.org/browse/OEUI-61)
## Summary:
This includes creating and styling the standard dosing form.


app/css/grid.scss
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+
+@for $i from 1 through 12 {
+    .pull-right-#{$i} {
+        @include push($i);
+    }
+}
+
+input[type=number]{
```
larrystone: I would suggest you use a class to define this property rather than making it global for all input fields of type number

flacode: Thank you, working on it now.

package.json
```diff
@@ -69,16 +74,18 @@
...
     "react-dom": "^16.2.0",
     "redux-logger": "^3.0.6",
     "redux-mock-store": "^1.5.1",
     "rimraf": "^2.5.2",
-    "style-loader": "^0.13.1",
+    "sass-loader": "^6.0.7",
+    "style-loader": "^0.13.2",
     "url-loader": "^0.5.7",
-    "webpack": "^1.12.13",
+    "webpack": "^1.15.0",
```
segunolalive: This is a legacy and perhaps unsupported version of webpack. Is there any reason for using this? Why not version 3 or 4(the latest)

app/css/grid.scss
```diff
@@ -0,0 +1,24 @@
...
+.grid-row {
+    @include row();
+}
+
+.button-row {
+    @include row();
+    margin-top: 1%;
+}
+
+@for $i from 1 through 12 {
```
segunolalive: It appears you're reimplementing a 12-column grid here. Why is that?

I also noticed the bootstrap scripts were commented out in `app/index.html`. Have you decided to drop bootstrap dependency?

webpack.sample.config.js
```diff
@@ -192,12 +192,15 @@ var webpackConfig = {
...
 			loader: 'url'
 		}, {
 			test: /\.html$/,
 			loader: 'html'
-		}],
+		}, {
+			test: /\.scss$/,
+			loaders: ["style-loader", "css-loader", "sass-loader"]
+	}],
 	},
```
zeze111: why make changes to the sample config and not the actual webpack.config file?

app/css/grid.scss
flacode: At first we had used bootstrap, but then we realised that openmrs has a style guide that we had not used. We decided to drop bootstrap for now and use the [styleguide](https://www.npmjs.com/package/@openmrs/style-referenceapplication). 
> It appears you're reimplementing a 12-column grid here. Why is that?
The style guide has a` grid.sccs` file that defines `mixins` for rows and columns which are just declarations and can not be used directly.

webpack.sample.config.js
flacode: The `webpack.config` file is included in the `gitignore`, so `webpack.sample.config` is provided as a generic file. This file is supposed to be copied into the `webpack.config` and then customize the paths according to your file system.

app/css/grid.scss
segunolalive: Using the *style-guide* does not necessarily exclude the need for bootstrap to provide consistent styling and widgets. Did using bootstrap have any negative effects on the styles?

package.json
flacode: Thank you @segunolalive for the observation. The version of the webpack is from the` owa-generator`. Migrating to webpack 3 or 4 is a good candidate for another task. Let me discuss this with my teammates. @dkayiwa @zeze111  @geofrocker @fuchodeveloper  @larrystone @Betty-Kebenei

app/index.html
```diff
@@ -9,19 +9,15 @@
...
-  <link rel="stylesheet" href="../../ms/uiframework/resource/referenceapplication/styles/referenceapplication.css"
-    type="text/css" />
-
   <link rel="stylesheet" href="../../ms/uiframework/resource/appui/styles/header.css" type="text/css"
   />
   <link rel="stylesheet" href="../../ms/uiframework/resource/uicommons/styles/styleguide/jquery-ui-1.9.2.custom.min.css" type="text/css"/>
-  <link rel="stylesheet" href="../../ms/uiframework/resource/orderentryui/styles/drugOrders.css" type="text/css"/>
-
+  
   <!-- Latest compiled and minified CSS -->
```
Betty-Kebenei: Why did you remove those css imports?

app/css/grid.scss
flacode: yes, we realised that our UI was not consistent with other interfaces in openmrs, it was a tradeoff for consistency.

webpack.sample.config.js
zeze111: I understand this, but since the changes are limited to just our project for now, shouldn't this then go into the ```webpack.config```

flacode: No, you will need to copy them into your `webpack.config`.

app/index.html
Betty-Kebenei: Okay, got it.

app/js/components/orderEntry/addForm/StandardDose.jsx
```diff
@@ -1,170 +1,154 @@
...
+          <label name="frequency">Frequency</label>
+          <input
+            id="frequency"
+            name="frequency"
+            list="orderFrequencies"
+            size="10"
+          />
+          <datalist id="orderFrequencies" >
+            <option value="Daily" />
+            <option value="Twice everyday" />
```
dkayiwa: Did we intentionally hard code these?

flacode: yes, you advised me to separate the validations from the UI.After this PR is merged, I will make another one with the dynamic data and its validations.

PR #29, opened 2018-04-17
OEUI-68: Optimising action's test

zeze111:
[OEUI-68: Optimising action's tests](https://issues.openmrs.org/browse/OEUI-68)

### SUMMARY:

This includes adding packages that have been called in each action's test file to the global setup, reducing data duplication in individual files.

### DIFF
--- a/tests/actions/drug.test.js
+++ b/tests/actions/drug.test.js
@@ -1,20 +1,12 @@
-import expect from 'expect';
-import configureMockStore from 'redux-mock-store';
-import thunk from 'redux-thunk';
-import moxios from 'moxios';
-
 import { searchDrug } from '../../app/js/actions/drug';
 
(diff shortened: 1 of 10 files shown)


tests/actions/drug.test.js
```diff
@@ -1,21 +1,9 @@
-import expect from 'expect';
-import configureMockStore from 'redux-mock-store';
-import thunk from 'redux-thunk';
-import moxios from 'moxios';
-
 import { searchDrug } from '../../app/js/actions/drug';
 
 import * as actionTypes from '../../app/js/actions/actionTypes';
```
flacode: @zeze111 , should not this also be importing specific modules?

zeze111: right, thanks

PR #30, opened 2018-04-17
OEUI-73: Optmising components' tests

zeze111:
[OEUI-73: Optimising components' tests](https://issues.openmrs.org/browse/OEUI-73)

### SUMMARY:

This includes adding packages that have been called in each component's test file to the global setup, reducing data duplication in individual files.


geofrocker:
@zeze111 work on the unsquashed commits

PR #31, opened 2018-04-18
OEUI-60: Confirm post/put drug order implementation (optimization)

Betty-Kebenei:
This includes:
- adding spaces to between values in the table
- adding a forgotten value (completeInstructions)


dkayiwa:
The link to the ticket does not look correct.

PR #33, opened 2018-04-18
OEUI-66: Fix bug in the drug search

larrystone:
## **JIRA TICKET NAME:**
[OEUI-66: Bug in the drug search input field.](https://issues.openmrs.org/browse/OEUI-66)

### SUMMARY:
Fix bug causing app redirecting to the homepage when user press `enter key` in drug search input field. Also fix bug with drug selection from drug options in search.


app/js/components/searchDrug/View.jsx
```diff
@@ -1,53 +1,81 @@
...
+          }}
+        />
+        {showLoading(loading)}
+        {showError(error)}
+        {noDrugFound()}
+      </span>
+      {renderResults(results)}
+    </div>
   );
 };
```
kwahalf: do we need this space?

app/js/components/searchDrug/index.jsx
```diff
@@ -3,10 +3,33 @@ import { debounce } from 'lodash';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
```
kwahalf: do we need this space?

app/js/actions/drug.js
```diff
@@ -1,24 +1,34 @@
...
   error,
 });
 
 export const searchDrugLoading = status => ({
-  type: actionTypes.SEARCH_DRUGS_LOADING,
+  type: SEARCH_DRUGS_LOADING,
   status,
 });
 
+export const selectDrugSuccess = drugUuid => ({
```
Betty-Kebenei: Why do you have two methods that look similar? `selectDrugSuccess` one passing drugs while the other drugUuid?

app/js/components/searchDrug/View.jsx
```diff
@@ -1,53 +1,81 @@
...
   );
 
-  const showLoading = () => (loading ? ' Searching...' : '');
-
-  const showError = () => (error ? error.data.message : '');
+  const noDrugFound = () => (
+    !loading && focused && results.length === 0 &&
+    <span className="preformatted-note">
+      <span className="emoji-sorry" role="img" aria-label="search"> 🔍 </span>
+      Please refine your search
```
Betty-Kebenei: Could this message be more descriptive?

app/js/reducers/drugSearchReducer.js
```diff
@@ -1,35 +1,45 @@
...
   drugs: [], selected: {}, error: null, loading: false,
 };
 
 export default (state = initialState, action) => {
   switch (action.type) {
-    case actionTypes.SEARCH_DRUGS_SUCCESS:
+    case SEARCH_DRUGS_SUCCESS:
       return {
         ...state,
         drugs: action.drugs,
```
kwahalf: since you are using es6 can ...actions work?

app/js/reducers/drugSearchReducer.js
```diff
@@ -1,35 +1,45 @@
...
-        selected: action.drugs && action.drugs.length === 1 ? action.drugs[0] : {},
+        selected: {},
         error: null,
```
kwahalf: why are we setting error to null and loading to false, while they were initially null and false

app/js/reducers/drugSearchReducer.js
```diff
@@ -1,35 +1,45 @@
...
         loading: false,
       };
-    case actionTypes.SEARCH_DRUGS_FAILURE:
+    case SEARCH_DRUGS_FAILURE:
       return {
         ...state,
         drugs: [],
         selected: {},
         error: action.error,
```
kwahalf: same here  try using ...actions

app/js/reducers/drugSearchReducer.js
```diff
@@ -1,35 +1,45 @@
...
         loading: false,
       };
-    case actionTypes.SEARCH_DRUGS_LOADING:
+    case SEARCH_DRUGS_LOADING:
       return {
         ...state,
         drugs: [],
         selected: {},
         error: null,
         loading: action.status,
```
kwahalf: same here

tests/components/searchDrug/View.test.jsx
```diff
@@ -39,20 +42,31 @@ describe('Component: searchDrug: View', () => {
...
     });
   });
+
+  describe('Select drug action', () => {
+    it('calls onOptionSelected()', () => {
+      props.focused = true;
+      const optionSelectedSpy = jest.spyOn(props, 'onOptionSelected');
+      const drugSearchInput = getComponent().find('.option').first();
+      drugSearchInput.simulate('click');
+
```
kwahalf: do we need this space inside an it block ?

app/js/actions/drug.js
larrystone: selectDrugSuccess action is called when a user selects a drug from the list of drugs, this make the selected drug available in the store for use by everyone. The one passing drugs is the one that makes a list of drugs available to the user based on his/her search parameter. I hope you understand

app/js/components/searchDrug/View.jsx
larrystone: I didn't want to use `Drug not found` because the logic here is to show a generic message that notifies the user when his/her search term is less than 3 characters and when there is no result of the search, so I felt that telling the user to `refine his search` will capture both scenario. I am however open to better options. Thanks

app/js/components/searchDrug/index.jsx
larrystone: I will work on this. thanks for the observation

app/js/reducers/drugSearchReducer.js
larrystone: That kinda looks promising, I will try it out

app/js/reducers/drugSearchReducer.js
larrystone: I will work on this and revert back. Thanks

tests/components/searchDrug/View.test.jsx
larrystone: I was considering using white spaces for code readability, but I think it's not necessary

app/js/reducers/drugSearchReducer.js
larrystone: Just tried this and it works great, thanks so much.

tests/components/searchDrug/View.test.jsx
larrystone: Thanks

larrystone:
@kwahalf I have implemented your feedback, thanks for your time and shared knowledge

app/js/actions/drug.js
```diff
@@ -1,24 +1,34 @@
...
-import * as actionTypes from './actionTypes';
+import {
+  SEARCH_DRUGS_LOADING,
+  SEARCH_DRUGS_FAILURE,
+  SEARCH_DRUGS_SUCCESS,
+  SELECT_DRUG,
+} from './actionTypes';
 
 const contextPath = window.location.href.split('/')[3];
 const apiBaseUrl = `/${contextPath}/ws/rest/v1`;
```
fuchodeveloper: These two lines have been abstracted to a separate file to avoid code duplication. Could you import the `apiBaseUrl` file instead of re-writing it again?

larrystone: okay, lemme do that then. Thanks

larrystone: Done!

PR #37, opened 2018-04-24
OEUI-81: Update the table display format for active and past orders

geofrocker:
## JIRA TICKET NAME:
[OEUI-81: Update the table display format for active and past orders](https://issues.openmrs.org/browse/OEUI-81)
### SUMMARY:
This task basically meant to implement the feedback got from the community about the fields that should appear in Active and Past order tables

### DIFF
--- a/app/js/components/orderEntry/PastOrders.jsx
+++ b/app/js/components/orderEntry/PastOrders.jsx
@@ -36,9 +36,7 @@ export class PastOrders extends React.Component {
             <thead>
               <tr>
                 <th className="w-145-px">Date</th>
-                <th className="w-120-px">Status</th>
                 <th>Details</th>
-                <th className="w-81-px">Actions</th>
               </tr>
             </thead>
             <tbody>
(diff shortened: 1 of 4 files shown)


app/js/components/orderEntry/PastOrders.jsx
```diff
@@ -36,17 +36,15 @@ export class PastOrders extends React.Component {
             <thead>
               <tr>
                 <th className="w-145-px">Date</th>
-                <th className="w-120-px">Status</th>
                 <th>Details</th>
```
fuchodeveloper: The past drug orders shown here has two table headers while the active drug orders table has three column headers. Would this implementation not result in an inconsistent UI?

Betty-Kebenei: While @fuchodeveloper  the last column for past drug orders had no meaning.

geofrocker: We agreed on this during the demo

PR #39, opened 2018-04-24
OEUI-83: Integration of a generic loader for active orders, past orders, add drug and Search for drug

fuchodeveloper:
## JIRA TICKET NAME

[OEUI-83: Integrate the generic loader for active orders, past orders, add drug and Search for drug](https://issues.openmrs.org/browse/OEUI-83)

## SUMMARY:
This ticket implements integration of a generic loader for active orders, past orders, add drug and Search for drug


larrystone:
I guess we have ourself some bugs here... please into the console here
screen shot 2018-04-26 at 1 30 17 pm

fuchodeveloper:
@larrystone How do I reproduce the above error?

larrystone:
@fuchodeveloper reloading the page and looking into the console should give that

fuchodeveloper:
Good to know.

app/js/actions/pastOrders.js
```diff
@@ -17,18 +17,18 @@ export const getPastOrdersFailure = error => ({
 });
 
 export const getPastOrders = (patientUuid, careSetting) => (dispatch) => {
-  dispatch(loading(true));
+  dispatch(loading('LOAD_PAST_ORDERS', true));
   return axios.get(`${apiBaseUrl}/order?careSetting=${careSetting}&patient=${patientUuid}&status=inactive&t=drugorder&v=full`)
```
kwahalf: I'll suggest that you create an axios instance with a base URL in some helper file then export it to get rid of appending ${apiBaseUrl} every time you make a call

fuchodeveloper: @kwahalf  Thanks for the suggestion. 
Have noted it.
Your feedback would be implemented as a separate task as this current PR is for implementation of the generic loader.

geofrocker: @kwahalf Thank you for the idea. This requires a change in all our previous tasks. A new ticket has been created for that [here](https://issues.openmrs.org/browse/OEUI-91)

PR #40, opened 2018-04-25
OEUI-85: implement actions and reducers to discontinue an order

zeze111:
[OEUI-85: Implement actions and reducers to discontinue an order](https://issues.openmrs.org/browse/OEUI-85)

### SUMMARY:

This is a sub-task to implement all actions and reducers related discontinuing/deleting an order


app/js/reducers/discontinueOrderReducer.js
```diff
@@ -0,0 +1,31 @@
+import {
+  DISCONTINUE_ORDER_FAILURE,
+  DISCONTINUE_ORDER_SUCCESS,
+  DISCONTINUE_ORDER,
```
Betty-Kebenei: @zeze111 I haven't seen added under actionTypes or used anywhere.

zeze111: thanks, must have missed it when I was removing it from my implementation

app/js/actions/discontinueOrder.js
```diff
@@ -0,0 +1,39 @@
...
+  type: DISCONTINUE_ORDER_SUCCESS,
+});
+
+export const discontinueOrderFailure = error => ({
+  type: DISCONTINUE_ORDER_FAILURE,
+  error,
+});
+
+
+export const discontinueOrder = (ordersPayload, patientUuid, careSetting) => (dispatch) => {
```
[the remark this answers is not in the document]
zeze111: Got it, thanks

app/js/reducers/discontinueOrderReducer.js
```diff
@@ -0,0 +1,30 @@
...
+];
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case DISCONTINUE_ORDER_SUCCESS:
+      return {
+        ...state,
+        activeOrders: state.activeOrders,
+        pastOrders: state.pastOrders,
+        loading: action.status,
```
larrystone: I can notice some code duplication here. state already contains activeOrders and pastOrders so I guess there is no need spreading them into the state. More like:
{
...state, loading: action.status
}

app/js/actions/discontinueOrder.js
```diff
@@ -0,0 +1,39 @@
...
+    .then(() => {
+      dispatch(
+        discontinueOrderSuccess(),
+        loading(DISCONTINUE_ORDER_SUCCESS, false),
+      );
+    }).catch((error) => {
+      dispatch(loading(DISCONTINUE_ORDER_FAILURE, false));
+
+      if (error.response) {
+        dispatch(discontinueOrderFailure(error));
```
larrystone: I feel sending the entire error object might be too verbose. Please consider `error.response` instead

app/js/reducers/discontinueOrderReducer.js
```diff
@@ -0,0 +1,30 @@
...
+      };
+    case DISCONTINUE_ORDER_FAILURE:
+      return {
+        ...state,
+        error: action.error,
+      };
+    default:
+      return state;
+  }
```
larrystone: I can't seem to find case for the Loading reducer... `LOADING_DISCONTINUE_ORDER_FAILURE`. Is this intentional?

zeze111: I don't have that as an action type because of the general loader, so if you look at the 2 switch cases, you'll see where they were used

PR #44, opened 2018-04-27
OEUI-46: Free text form styling and validation

Betty-Kebenei:
# **JIRA TICKET NAME:**

[OEUI-46: Free text form styling and validation](https://issues.openmrs.org/projects/OEUI/issues/OEUI-46)

## **SUMMARY:**
- Style the free text form to be consistent with the standard
dosing form.
- Display the dispense fields only if the caresetting is for
outpatient.
- Validate the free text form


app/js/components/orderEntry/addForm/AddForm.jsx
```diff
@@ -217,19 +217,22 @@ export class AddForm extends React.Component {
 
     if (Object.values(this.state.fieldErrors).includes(true)) return true;
```
larrystone: Have you tried `return Object.values(this.state.fieldErrors).includes(true)`?
I think it will give us the same result

app/js/components/orderEntry/addForm/AddForm.jsx
```diff
@@ -217,19 +217,22 @@ export class AddForm extends React.Component {
...
 
-    else if (duration && !durationUnit) return true;
-
-    else if ((this.props.careSetting.display === 'Outpatient') &&
+    else if (this.state.formType === 'Standard Dosage') {
+      if (duration && !durationUnit) return true;
```
larrystone: Same here...Basically, we can remove all the switch statement and replace with return statements.

app/js/components/orderEntry/addForm/AddForm.jsx
```diff
@@ -217,19 +217,22 @@ export class AddForm extends React.Component {
...
-
-    else if ((this.props.careSetting.display === 'Inpatient') &&
-    !(dose && dosingUnit && frequency && route)) return true;
-
+      else if ((this.props.careSetting.display === 'Inpatient') &&
+      !(dose && dosingUnit && frequency && route)) return true;
+    } else if ((this.state.formType === 'Free Text')) {
+      if ((this.props.careSetting.display === 'Outpatient') &&
+      !(dispensingQuantity &&
+        dispensingUnit)) return true;
```
larrystone: Please look into these lines, I guess it can be improved....

app/js/components/orderEntry/addForm/AddForm.jsx
Betty-Kebenei: If I use the return statement there, all the other code below will not be reachable.

larrystone: Well, I think your code will also make the lines after it unreachable when the condition is true. return statements generally block the execution of the statements after them

Betty-Kebenei: @larrystone With the way this is implemented it is working okay because of the condition `if` but using what you have suggested makes the code unreachable because the condition is removed.

app/js/components/orderEntry/addForm/AddForm.jsx
```diff
@@ -215,23 +215,30 @@ export class AddForm extends React.Component {
...
+      )) return true;
+      else if ((this.props.careSetting.display === 'Inpatient') &&
+      !(dose && dosingUnit && frequency && route)) return true;
+    } else if (this.state.formType === 'Free Text') {
+      if ((this.props.careSetting.display === 'Outpatient') &&
+      !(drugInstructions && dispensingQuantity && dispensingUnit)) return true;
+      else if ((this.props.careSetting.display === 'Inpatient') &&
+      !(drugInstructions)) return true;
+    }
     return false;
```
mogoodrich: Generally by OpenMRS convention we also put brackets around if and else blocks... 

https://stackoverflow.com/questions/2125066/is-it-bad-practice-to-use-an-if-statement-without-brackets

Betty-Kebenei: @mogoodrich Thank you, will fix that

PR #46, opened 2018-04-30
OEUI-92: Consume API to an fetch encounter role

Betty-Kebenei:
# **JIRA TICKET NAME:**

[OEUI-92: Consume API to an fetch encounter role](https://issues.openmrs.org/browse/OEUI-92)

## **SUMMARY:**
This includes consuming the API to fetch an encounter role which is needed when making a drug order. Encounter Role states the role of the provider of an order.


app/js/reducers/encounterRoleReducer.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const encounterRole = (state = initialState, action) => {
+  switch (action.type) {
+    case FETCH_ENCOUNTER_ROLE_LOADING:
+      return {
+        ...state,
+        isLoading: action.status,
+        encounterRole: {},
+        error: null,
```
larrystone: Lines 19 and 20 are kinda unnecessary because those values are already available as default in the initialState, so I guess it's safe to remove them. Thanks

app/js/reducers/encounterRoleReducer.js
```diff
@@ -0,0 +1,41 @@
...
+      };
+    case FETCH_ENCOUNTER_ROLE_SUCCESS:
+      return {
+        ...state,
+        isLoading: false,
```
larrystone: Same here, isLoading is not necessary as it is already set to false on line 8

app/js/reducers/encounterRoleReducer.js
```diff
@@ -0,0 +1,41 @@
...
+        encounterRole: action.encounterRole,
+        error: null,
```
larrystone: Same with error here

app/js/reducers/encounterRoleReducer.js
```diff
@@ -0,0 +1,41 @@
...
+      };
+    case FETCH_ENCOUNTER_ROLE_FAILURE:
+      return {
+        ...state,
+        isLoading: false,
+        encounterRole: {},
+        error: action.error,
```
larrystone: Same here, isLoading and encounterRole can be safely removed

PR #63, opened 2018-05-11
OEUI-100: Checking the existence of encounter role

larrystone:
## **JIRA TICKET NAME:**
[OEUI-100: Checking the existence of encounter type.](https://issues.openmrs.org/browse/OEUI-100)

### SUMMARY:
This entails checking whether the encounter role whose name was set as a global property exists. If this encounter role does not exist then a user should get an error message on loading the drug order page.

This handle logic for validating proper and complete setup of the encounterRole configurations and notifies the user if not.
i.e `if return axiosInstance.get(encounterrole?q=${value})` is not successful, then an error is displayed like for *Configuration for encounterRole is incomplete*

### DIFF
--- a/app/js/reducers/settingEncounterRoleReducer.js
+++ b/app/js/reducers/settingEncounterRoleReducer.js
@@ -17,8 +17,6 @@ const settingEncounterRoleReducer = (
       return { ...state, isLoading: action.status };
     case SETTING_ENCOUNTER_ROLE_FAILURE:
       return { ...state, roleError: action.error };
-    case NETWORK_ERROR:
-      return { ...state, roleError: action.error };
     default:
       return state;
   }
--- a/tests/actions/settingEncounterRole.test.js
+++ b/tests/actions/settingEncounterRole.test.js
@@ -9,7 +9,11 @@ import {
     SETTING_ENCOUNTER_ROLE_LOADING,
 } from '../../app/js/actions/actionTypes';
 
-describe ('Get the encounterRole configuration actions', () => {
+jest.mock('../../app/js/actions/encounterRole', () => ({
+    fetchEncounterRole: (value) => jest.fn()
+}));
+
+describe('Get the encounterRole configuration actions', () => {
     it('should create an action to store fetched configuration', () => {
         const configuration = 'order entry';
         const expectedAction = {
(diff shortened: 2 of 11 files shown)


fuchodeveloper:
Please look into the failing tests.

larrystone:
Thanks @hadijahkyampeire and @fuchodeveloper. I have fixed the test

tests/actions/settingEncounterRole.test.js
```diff
@@ -45,8 +46,33 @@ describe ('Get the encounterRole configuration actions', () => {
...
+            const store = mockStore({ setting: {}});
+
+            return store.dispatch(getSettingEncounterRole()).then(() => {
+                const dispatchedActions = store.getActions();
+                const dispatchedActionTypes = dispatchedActions.map(action => action.type);
+                expect(dispatchedActionTypes).toEqual(expectedActions);
+            });
+        });
+
+        it ('should dispatch `SETTING_ENCOUNTER_ROLE_SUCCESS` with empty results', () => {
```
flacode: @larrystone, I dont see `SETTING_ENCOUNTER_ROLE_SUCCESS` in the expected actions.

tests/actions/settingEncounterRole.test.js
```diff
@@ -7,6 +7,7 @@ import {
     SETTING_ENCOUNTER_ROLE_SUCCESS,
     SETTING_ENCOUNTER_ROLE_FAILURE,
     SETTING_ENCOUNTER_ROLE_LOADING,
+    FETCH_ENCOUNTER_ROLE_LOADING,
```
flacode: @larrystone, in my opinion, testing `FETCH_ENCOUNTER_ROLE` in this file makes this an integrated test instead of a unit test. You could think of mocking the fetchEncounterole action.

tests/actions/settingEncounterRole.test.js
larrystone: Oh, my bad, the implementation logic is that it should throw a failure if an empty result was returned when fetching SETTING_ENCOUNTER_ROLE. I will update the test case description accordingly. Thanks

tests/actions/settingEncounterRole.test.js
larrystone: Thanks for bringing this up. I have fixed this

app/js/reducers/settingEncounterRoleReducer.js
```diff
@@ -17,8 +17,6 @@ const settingEncounterRoleReducer = (
       return { ...state, isLoading: action.status };
     case SETTING_ENCOUNTER_ROLE_FAILURE:
       return { ...state, roleError: action.error };
-    case NETWORK_ERROR:
```
dkayiwa: Do you mind explaining a bit how removing the network error lines is related to the work for this ticket?

larrystone: Actually, it doesn't directly have anything to with this ticket but having to refactor the test for this file made me realise that this line was not been used anywhere and therefore unnecessary, hence the need to remove it.

PR #66, opened 2018-05-14
OEUI-86: Implement updates to POST discontinuing an order

zeze111:
[OEUI-86: Implement updates to POST discontinuing an order](https://issues.openmrs.org/browse/OEUI-86)

### SUMMARY:
This PR includes additional updates to current implementation of posting a drug order to allow an order to be discontinued


app/js/components/orderEntry/ActiveOrders.jsx
```diff
@@ -37,33 +37,42 @@ export class ActiveOrders extends React.Component {
...
         uuid,
+        careSetting: careSetting.uuid,
         drugName: drug.display,
         action: 'DISCONTINUE',
         dose,
-        dosingUnit,
-        frequency,
-        route,
+        drug: drug.uuid,
+        dosingUnit: (doseUnits && doseUnits.display) || doseUnits,
```
humanityjs: Hey. What is this line suppose to do? Just curious.

app/js/components/orderEntry/ActiveOrders.jsx
```diff
@@ -37,33 +37,42 @@ export class ActiveOrders extends React.Component {
...
+        frequency: (frequency && frequency.display) || frequency,
```
humanityjs: Same here

app/js/components/orderEntry/ActiveOrders.jsx
zeze111: it checks that there is a valid "value" in "doseUnits", if so then return "doseUnits.display" or else return whatever value is in "doseUnits", which can either be "null" or empty string " ".

app/js/components/orderEntry/ActiveOrders.jsx
zeze111: Same as above

PR #71, opened 2018-05-15
OEUI-99: Update action test files to improve the codebase test coverage

flacode:
### JIRA TICKET NAME 
[OEUI-99: Update action test files to improve the codebase test coverage](https://issues.openmrs.org/browse/OEUI-99)

## Summary:
- Add missing test files for action creators.
- Updating existing test files to improve the test coverage for action creators.


app/js/actions/careSetting.js
```diff
@@ -6,17 +6,17 @@ import axiosInstance from '../config';
 import networkError from './networkError';
 import loading from './loading';
 
-const fetchPatientCareSettingActionCreator = patientCareSetting => ({
+export const fetchPatientCareSettingActionCreator = patientCareSetting => ({
```
hadijahkyampeire: Looks good, though I have a recommendation, what if you do all exports once at the end of the file, imagine when many action creators are added and the export word is repeating itself.

flacode: @hadijahkyampeire, thank you for your observation. I think it is a matter of preference and consistency with the rest of the files in the project.

hadijahkyampeire: ok

tests/actions/orderEntryActions.test.js
```diff
@@ -61,14 +61,42 @@ describe ('Order entry configuration actions', () => {
             });
         });
 
-        it ('should dispatch `FETCH_ORDER_CONFIG_FAILURE` after an error', () => {
+        it ('should dispatch `FETCH_ORDER_CONFIG_FAILURE` if status is not 200', () => {
             moxios.wait(() => {
                 const request = moxios.requests.mostRecent();
                 request.respondWith({
+                    status: 204,
```
fuchodeveloper: You have a status of `204` but dispatching a failure.
Could you have a look at this resource for the 204 status code - [link](https://httpstatuses.com/204)

tests/actions/orderEntryActions.test.js
```diff
@@ -61,14 +61,42 @@ describe ('Order entry configuration actions', () => {
...
+                expect(dispatchedActionTypes).toEqual(expectedActions);
+            });
+        });
+
+
+        it ('should dispatch `FETCH_ORDER_CONFIG_FAILURE` after an error', () => {
+            moxios.wait(() => {
+                const request = moxios.requests.mostRecent();
+                request.reject({
                     status: 401,
```
fuchodeveloper: Here you have the same error dispatch but a status of `401`

tests/actions/pastOrders.test.js
```diff
@@ -1,36 +1,76 @@
...
+    };
+    const store = mockStore({});
+    await store.dispatch(getPastOrders(limit, startIndex, patientUuid, careSetting))
+      .then(() => {
+        expect(store.getActions()).toEqual(expect.arrayContaining([expect.objectContaining(expectedActions)]));
+      });
+    done();
+  });
+
+  it('should respond to an error', async (done) => {
```
fuchodeveloper: Would be helpful if you specify the error expected.

package.json
```diff
@@ -18,6 +18,7 @@
     "deep-freeze": "0.0.1",
     "eslint-config-airbnb-base": "^12.1.0",
     "lodash": "^4.17.5",
+    "npm": "^6.0.0",
```
humanityjs: Why do you have npm as a dependency?

fuchodeveloper:
Good job @flacode 
Please look into the comments.

tests/actions/orderEntryActions.test.js
flacode: The action creator throws an error when the status code is not equal to 200. So since 204 is for the successful request without content, I decided to use it in my tests.

tests/actions/orderEntryActions.test.js
flacode: 401 is the status code for unauthorized and axios catches this as an error explaining the error dispatch.

tests/actions/pastOrders.test.js
flacode: noted.

tests/actions/orderEntryActions.test.js
fuchodeveloper: Okay.
If the expected action is an error, would it not be better to use a status code that more clearly shows this? 
`20*` status codes might not be the most appropriate in this situation.

flacode: The expected action corresponds to a success code that is not 200. The error is thrown by the action, not the API.

tests/actions/orderEntryActions.test.js
fuchodeveloper: If this error occurs due to unauthorized access, could you indicate this in the `it` description?
This would help make it clearer.

tests/actions/orderEntryActions.test.js
fuchodeveloper: Just for more clarity.
A `204` status code means that there was not a failure in the action dispatch, rather it was successful but there is no content. This is quite different from what the `it` description in this test specifies.

humanityjs: @flacode . I guess what @fuchodeveloper is saying is that a normal `axios` request will resolve a response with a status code of `204`. It won't be rejected and hence it won't go to the `catch` block. Your implementation might be right if you don't have a say over the status code coming from the API and all of that. But you should use a more precise status code if you can. 

And if you ask me, in this case (since I'm fetching data), I'll return a status `200` and an empty array and handle that in my `actionCreator`.

package.json
```diff
@@ -54,9 +54,9 @@
     "enzyme-to-json": "^3.3.3",
     "eslint": "^4.19.1",
     "eslint-config-airbnb": "^16.1.0",
-    "eslint-plugin-import": "^2.9.0",
+    "eslint-plugin-import": "^2.11.0",
     "eslint-plugin-jsx-a11y": "^6.0.3",
-    "eslint-plugin-react": "^7.7.0",
+    "eslint-plugin-react": "^7.8.2",
```
zeze111: why did you upgrade the eslint versions?

tests/actions/orderEntryActions.test.js
flacode: @humanityjs, I agree with you. But how else would I test this async action? 
`return axiosInstance.get(`orderentryconfig?v=custom:(uuid,display)`)
    .then((response) => {
      if (response.status !== 200) {
        throw Error(response.statusText);
      }
      dispatch(loading('FETCH_ORDER_CONFIG', false));
      return response;
    })
    .then(response => dispatch(fetchConfigurationsSuccess(response.data)))
    .catch((error) => {
      dispatch(loading('FETCH_ORDER_CONFIG', false));
      dispatch(fetchConfigurationsFailure(error));
    });`

tests/actions/orderEntryActions.test.js
flacode: noted, thank you.

package.json
flacode: That from running npm install as I was setting up on a new machine.

tests/actions/pastOrders.test.js
```diff
@@ -1,36 +1,76 @@
...
-  LOAD_PAST_ORDERS_FAILURE
+  LOAD_PAST_ORDERS_FAILURE,
+  NETWORK_ERROR
 } from '../../app/js/actions/actionTypes';
 import { getPastOrders } from '../../app/js/actions/pastOrders';
 
-window.location = locationMock;
-
-const uuid = '761d8414-6cb4-4456-be4f-232e6f819e06';
+const limit = 10;
```
Betty-Kebenei: @flacode I would suggest you make use of the mockData file

flacode: @Betty-Kebenei, could you please help me understand this.

dkayiwa: @Betty-Kebenei do you mind helping her understand that? 😊

PR #84, opened 2018-05-22
OEUI-128: Correct tooltips with incorrect labels

fuchodeveloper:
## JIRA TICKET NAME

[OEUI-128:  Correct tooltips with incorrect labels](https://issues.openmrs.org/browse/OEUI-128)

## SUMMARY:
The wording for some tooltips does not correspond to what the buttons actually do.

Tooltip for edit draft drug order showing `delete` instead of `edit`.
Tooltip for discontinue draft drug order showing `delete` instead of `discontinue`
Plus any other tooltips that need to be corrected.

### DIFF
--- a/app/js/components/orderEntry/addForm/DraftDataTable.jsx
+++ b/app/js/components/orderEntry/addForm/DraftDataTable.jsx
@@ -187,7 +187,7 @@ export class DraftDataTable extends React.Component {
             <i className="icon-edit" title="Edit" />
           </a>
           <a id="discard-draft-order" href="#" onClick={() => handleDiscardOneOrder(order)}>
-            <i className="icon-remove" title="Delete" />
+            <i className="icon-remove" title="Discard" />
           </a>
         </td>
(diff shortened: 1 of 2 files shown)


app/js/components/orderEntry/addForm/DraftDataTable.jsx
```diff
@@ -184,7 +184,7 @@ export class DraftDataTable extends React.Component {
             <i className="icon-edit" title="Edit" />
           </a>
           <a id="discard-draft-order" href="#" onClick={() => handleDiscardOneOrder(order)}>
-            <i className="icon-remove" title="Delete" />
+            <i className="icon-remove" title="Discontinue" />
```
Betty-Kebenei: You should use discard for this part

fuchodeveloper: Would have to consult the team to get more feedback on this.

PR #87, opened 2018-05-23
OEUI-126: Clear and close drug forms and unsaved draft orders table on switching between Inpatient and Outpatient tabs

Betty-Kebenei:
# **JIRA TICKET NAME:**

[OEUI-126: Clear and close drug forms and unsaved draft orders table on switching between Inpatient and Outpatient tabs](https://issues.openmrs.org/browse/OEUI-126)

## **SUMMARY:**

The Inpatient and Outpatient tabs on the Order Entry UI seem to share the same entry form.  By this I mean it is possible to switch between the two tabs as the user is filling out the form.  It is saved to Inpatient or Outpatient based on which tab is chosen when the user clicks save.  This can cause some confusion as there are different fields for the two tabs: e.g. The user may start in the Outpatient Tab, which includes dispensing units and then switch over to Inpatient and save, which would result in an Inpatient order with dispensing units filled out.  

It is also a little confusing during editing.  If you edit an Inpatient order, you can switch over to Outpatient.  If you try to save it like this, you will get a Toast message with the error "the care setting does not match that of the previous order".  It's good that it does this, but the user should not be allowed to attempt to edit an Inpatient order on the Outpatient tab.

 

It seems these 2 tabs should be separate and draft/edited orders should not be shared between them.

### DIFF
--- a/app/js/components/orderEntry/SearchAndAddOrder.jsx
+++ b/app/js/components/orderEntry/SearchAndAddOrder.jsx
@@ -98,6 +98,12 @@ export class SearchAndAddOrder extends React.Component {
     });
   }
 
+  closeFormsOnTabChange = () => {
+    this.clearSearchField();
+    this.props.selectDrugSuccess();
+    this.props.deleteAllDraftOrders();
+  }
+
   renderSearchDrug = () => (
     this.state.editDrugName ?
       <p className="revise-order-padding">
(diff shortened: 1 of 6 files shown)


app/js/components/orderEntry/SearchAndAddOrder.jsx
```diff
@@ -91,6 +91,11 @@ export class SearchAndAddOrder extends React.Component {
     });
   }
 
+  closeFormsOnTabChange = () => {
+    this.props.selectDrugSuccess('');
```
paradisekelechi: Why the use of an empty string as a parameter? Why not using a null value, i.e not sending anything at all?

dkayiwa:
Can you edit the ticket and pull request title to reflect a summary of that?

PR #89, opened 2018-05-23
OEUI-117: Drop downs options should only be filtered by the first letter 

flacode:
### JIRA TICKET NAME 
[OEUI-117: Drop downs options should only be filtered by the first letter](https://issues.openmrs.org/browse/OEUI-117)

## Summary:
Currently on the dropdowns in the Order Entry UI when the user types any letter or combination of letters in any of combo boxes, the choices are filtered based on those letters appearing anywhere in the choices.    It would be better if the results were filtered based on those letters appearing in the beginning of the choices.  e.g. if a user types "ta" in the units box, they probably want tablet or tablespoon, but not container.

### DIFF
--- a/app/js/components/orderEntry/addForm/AddForm.jsx
+++ b/app/js/components/orderEntry/addForm/AddForm.jsx
@@ -35,6 +35,13 @@ export class AddForm extends React.Component {
     formType: 'Standard Dosage',
     dosingType: '',
     activeTabIndex: 0,
+    options: {
+      dosingUnit: [],
+      frequency: [],
+      route: [],
+      durationUnit: [],
+      dispensingUnit: [],
+    },
   };
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


app/js/components/orderEntry/addForm/AddForm.jsx
```diff
@@ -148,13 +156,39 @@ export class AddForm extends React.Component {
...
     }); }
   }
 
   handleChange = (event) => {
+    const { name, value } = event.target;
+    const {
+      drugDosingUnits, orderFrequencies, drugRoutes, durationUnits, drugDispensingUnits,
+    } = this.props.allConfigurations;
+    let options;
+    if (value && name === 'dosingUnit') {
```
paradisekelechi: Noticing that the value is tested, it would be better practice to nest other conditions inside the condition that tests value. Thus, we would have something like this: 

```
if(value){
    switch (name) {}
}else{
}
```
With the above, you'd handle more conditions and would have a neater code.
@flacode

flacode: @paradisekelechi ,  I wanted to use a switch statement, but I noticed that there is no default case therefore if else would be a better option.

flacode: thank you for the resource

PR #106, opened 2018-06-10
OEUI-163: Make dateTimeFormat key consistent with others

larrystone:
## **JIRA TICKET NAME:**
[OEUI-163: Make dateTimeFormat key consistent with others](https://issues.openmrs.org/browse/OEUI-163)

### SUMMARY:
The global property key for setting up date and Time is orderentryowa.dateAndTimeFormat which is not consistent with the other global properties e.g order.encounterType.

This ticket is to fix this inconsistency by changing the dateTime key to order.dateTimeFormat


dkayiwa:
orderentryowa. is more representative of the intended use case than order.

larrystone:
@dkayiwa is it okay to revert back to the existing `orderentryowa.dateAndTimeFormat` key then?

PR #111, opened 2018-07-11
OEUI-169: Consume Endpoint to create Lab Order

Efosaok:
#### JIRA TICKET NAME
[Consume the endpoint for adding a Lab order](https://issues.openmrs.org/browse/OEUI-169?jql=project%20%3D%20OEUI%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)

#### Summary
- Consume resource to create a lab order.
- display the appropriate response from the server in the view.


app/js/actions/createLabOrder.js
```diff
@@ -0,0 +1,41 @@
...
+
+const createLabOrder = labOrderData => async (dispatch) => {
+  try {
+    dispatch(loading(CREATE_LAB_ORDER, true));
+    const requestPayload = {
+      name: randomString.generate(10),
+      description: JSON.stringify(labOrderData),
+      javaClassName: 'org.openmrs.Laborder',
+    };
+    const response = await axiosInstance.post('/ordertype', requestPayload);
```
IAMOTZ: I see you are saving the Lab Orders into the `orderType` endpoint. I am not sure this a good idea and that is because:

- The main purpose of this endpoint is to store the order types and not the orders themselves. If there is a functionality that depends on the response from this endpoint, that functionality might begin to break.
- What happens when we need to fetch the Lab Orders for display, even though the lab orders are saved there, we would be fetching the order-types and not really the Lab Orders, consequently, displaying the lab orders would be very difficult.

I am aware that you are having a blocker on saving the Lab Orders into the `/order` endpoint as you have posted [here](https://talk.openmrs.org/t/issue-making-a-post-request-to-create-a-test-order-order-type/18908), I suggest you work toward fixing the blocker as opposed to using this approach.

app/js/actions/createLabOrder.js
```diff
@@ -0,0 +1,41 @@
...
+  type: CREATE_LAB_ORDER_SUCCESS,
+  data,
+});
+
+const createLabOrder = labOrderData => async (dispatch) => {
+  try {
+    dispatch(loading(CREATE_LAB_ORDER, true));
+    const requestPayload = {
+      name: randomString.generate(10),
+      description: JSON.stringify(labOrderData),
```
IAMOTZ: I see you are using a random string as the name of the lab order, I am not sure this is supposed to be.
If it is supposed to be, please provide a reason for that.
If it is not supposed to be, I believe you should be able to get the name of the Lab Order from the details provided by the user.

app/js/components/labOrderEntry/LabEntryForm.jsx
```diff
@@ -67,10 +123,7 @@ export class LabEntryForm extends React.Component {
   }
 
   handleSubmit = () => {
-    /* *
-     * this should contain data to be submitted
-     * * */
-    this.handleCancel();
+    this.props.createLabOrder(this.state);
```
IAMOTZ: I see you are passing the whole `state` of this component to the `createLabOrder` action. I don't think this is a good idea.
From the implementation of the `createLabOrder` action you created, you stringify the argument it receives and save it straight away into the database. 
Calling the `createLabOrder` action like this means you would end up stringifying the whole `state` of this component and save it into the database.
The `state` of this component contains a lot that are not even part of the lab order information, I think you need to do some refactoring here.

app/js/reducers/createLabOrderReducer.js
```diff
@@ -0,0 +1,40 @@
...
+    case 'CREATE_LAB_ORDER_FAILURE': {
+      return {
+        ...state,
+        errorMessage: action.error,
+        status: {
+          error: true,
+          added: false,
+        },
+      };
+    }
```
IAMOTZ: I see you are specifying the action types as strings in this reducer.
Though this would work fine, there is an `actionTypes.js` file that I believe should hold all the action types as constants for easy reusability.
I think adding the actionTypes to that file would be a nice refactoring.

Efosaok: I get your point, if you look into the action creator for this PR, I imported the action types from the action types file, It is just my personal preference to not do that when I am working on reducers and on tests, everything in action remains in action.

app/js/components/labOrderEntry/LabEntryForm.jsx
Efosaok: ok, I'll look into that

app/js/actions/createLabOrder.js
Efosaok: the bug is coming from the API itself, and this is just an alternative until the issue with the order resource endpoint is rectified.

app/js/actions/createLabOrder.js
Efosaok: To save into the order type resource, the name field has to be unique, so all the data relating to the lab order is in the description field and the random string simply serves as a placeholder for a name in the database,
again, this is just an alternative until the bug with the order resource endpoint is fixed.

app/js/reducers/createLabOrderReducer.js
IAMOTZ: Awesome 👍 
Thanks for the clarification.

app/js/actions/createLabOrder.js
IAMOTZ: Please check this: https://talk.openmrs.org/t/issue-making-a-post-request-to-create-a-test-order-order-type/18908/7?u=tunmi
Mark Goodrich is asking you to create a ticket on REST WS module to resolve the bug you are facing. He also commented on the approach on saving the Order through the `orderType` endpoint not being a suitable one.
Please follow up with the bug fix.

app/js/actions/createLabOrder.js
IAMOTZ: If the reason you are using a random string is the bug you are facing with storing data through the `order` endpoint, please follow up with the discussion in this post on how to go about fixing the bug:  https://talk.openmrs.org/t/issue-making-a-post-request-to-create-a-test-order-order-type/18908/7?u=tunmi

dkayiwa:
Some merge conflicts here.

Efosaok:
This branch cannot be merged until the blocker is resolved, I'll resolve the merge conflicts as soon as I fix the blocker

dkayiwa:
@Efosaok which blocker are you fixing?

Efosaok:
The blocker on how to add a resource to the order table, this branch contains an alternative solution to that blocker, but it is not sustainable. so we'll be trying to use the approach Mark suggested yesterday to get past the blocker.

dkayiwa:
@Efosaok when will you try that other approach?

Efosaok:
Today

dkayiwa:
@Efosaok meaning i should expect a pull request from you by end of today?

Efosaok:
I am simply trying out an approach, not sure if it will work or not, but if it works, then this PR should be good to go by tomorrow, the endpoint is the only issue here.

dkayiwa:
Ping me for help as soon as you get blocked.

Efosaok:
ok, thanks

app/js/actions/createLabOrder.js
```diff
@@ -0,0 +1,185 @@
...
+});
+
+const createLabOrder = labOrderData => async (dispatch) => {
+  try {
+    dispatch(loading(CREATE_LAB_ORDER, true));
+    const requestPayload = {
+      encounterProviders: [
+        {
+          encounterRole: '240b26f9-dd88-4172-823d-4a8bfeb7841f',
+          provider: 'f9badd80-ab76-11e2-9e96-0800200c9a66',
```
mogoodrich: These should come from variables, like they do for drug order... if you are hardcoding these for testing, that's fine, but it may not work on another users system:

https://github.com/openmrs/openmrs-owa-orderentry/blob/master/app/js/components/drugOrderEntry/addForm/DraftDataTable.jsx#L93

app/js/actions/createLabOrder.js
```diff
@@ -0,0 +1,185 @@
...
+        },
+      ],
+      encounterType: '7b68d557-85ef-4fc8-b767-4fa4f5eb5c23',
+      location: {
```
mogoodrich: This should just be a uuid that references a current location, and not create a whole new location.

app/js/actions/createLabOrder.js
```diff
@@ -0,0 +1,185 @@
...
+          },
+          {
+            rel: 'full',
+            uri:
+              'http://localhost:8081/openmrs-standalone/ws/rest/v1/location/b1a8b05e-3542-4037-bbd3-998ee9c40574?v=full',
+          },
+        ],
+        resourceVersion: '2.0',
+      },
+      orders: [
```
mogoodrich: I think I see the issue here... test drugs have different parameters that drug orders... let me collect what they are and will post on the Talk post.

dkayiwa:
@Efosaok can you look into these merge conflicts?

PR #113, opened 2018-07-15
OEUI-172: Remove error messages when running test suites

andela-jomadoye:
**JIRA TICKET NAME**
[OEUI-172](https://issues.openmrs.org/browse/OEUI-172) Remove error messages in test

**Summary**

- This PR fixes most of the warning messages being displayed when running the test cases.

### DIFF
--- a/tests/components/DrugOrderEntry/index.test.jsx
+++ b/tests/components/DrugOrderEntry/index.test.jsx
@@ -6,16 +6,18 @@ import ConnectedDraftTable from '../../../app/js/components/drugOrderEntry/addFo
 const { order } = mockData;
 const props = {
   outpatientCareSetting:{
-    uuid:{}
+    uuid: '',
+    display: '',
   },
   inpatientCareSetting: {
-    uuid: {}
+    uuid: '',
+    display: '',
   },
   getPastOrders: jest.fn(),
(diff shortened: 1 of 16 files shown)


tests/components/DrugOrderEntry/index.test.jsx
```diff
@@ -188,22 +192,25 @@ describe('behaviour when the length of the unsaved draft orders table is more th
...
+        display: '',
       },
       getPastOrders: jest.fn(),
       draftOrders: [
         {dose: '', doseUnits: ''}
       ],
       setOrderAction: jest.fn(),
       location:{
-        search:()=>{}
+        search: '',
```
topseySuave: But why is it changed to a string when `location.search()` is supposed to be a function

andela-jomadoye: I believe the `location.search` property is a string as defined [HERE](https://www.w3schools.com/jsref/prop_loc_search.asp).
Also the Proptypes of the component defines it as 
```
location: PropTypes.shape({
    search: PropTypes.string,
  }).isRequired,
```

topseySuave: Ok cool.

PR #121, opened 2018-07-30
OEUI-196: Get lab orderables on initial page load

topseySuave:
## JIRA TICKET NAME
[OEUI-196](https://issues.openmrs.org/browse/OEUI-196) Get lab orderables on initial page load

## SUMMARY
This creates an action for getting Lab Orderables from the API endpoint.


tests/reducers/labOrderReducers/labOrderableReducer.test.js
```diff
@@ -0,0 +1,44 @@
...
+      labOrderables: [{}, {}, {}],
+      loading: false,
+      error: {
+        message: 'Error in connections',
+      },
+    };
+    const actualState = labOrderableReducer(initialState, fetchLabOrderablesFailure(error));
+    expect(actualState).toEqual(expectedState);
+  });
+});
```
jomadoye: Small EOL error here :-)

app/js/actions/actionTypes.js
```diff
@@ -47,3 +47,6 @@ export const DELETE_ALL_DRAFT_LAB_ORDERS = 'DELETE_ALL_DRAFT_LAB_ORDERS';
 export const SET_ORDER_ACTION = 'SET_ORDER_ACTION';
 export const GET_DATE_SUCCESS = 'GET_DATE_SUCCESS';
 export const GET_DATE_FAILURE = 'GET_DATE_FAILURE';
+export const FETCH_LAB_ORDERABLE_LOADING = 'FETCH_LAB_ORDERABLE_LOADING';
```
jomadoye: Should this be `FETCH_LAB_ORDERABLES_LOADING` (the orderable is plural).
What do you think? and the same for success and failure.

app/js/actions/labOrders/labOrderableAction.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+export const fetchLabOrderablesFailure = error => ({
+  type: types.FETCH_LAB_ORDERABLE_FAILURE,
+  error,
+});
+
+/**
+ * @function
+ * @returns {Object} - Object consisting lab orderables from API resource
```
IAMOTZ: A little bit more effort is required in this JSDoc documentation 🙂.
The documentation states that the `function` below is returning an`object` whereas it is actually returning  a `function`.

app/js/actions/labOrders/labOrderableAction.js
```diff
@@ -0,0 +1,35 @@
...
+  return axiosInstance.get('/concept/da006137-88ca-4d11-ae58-8b4b439afdd6?v=full')
+    .then((response) => {
+      if (response.status !== 200) {
+        throw Error(response.statusText);
+      }
+      dispatch(loading('FETCH_LAB_ORDERABLE', false));
+      return response;
+    })
+    .then(({ data }) => dispatch(fetchLabOrderablesSuccess(data)))
+    .catch((error) => {
```
IAMOTZ: You can put the `dispatch(fetchLabOrderablesSuccess(data))` in the first `then` block consequently you don't need a second `then` block.

app/js/reducers/initialState.js
```diff
@@ -108,4 +108,8 @@ export default {
     dateFormat: '',
     error: null,
   },
+  labOrderables: {
+    error: false,
+    loading: false,
+  },
```
IAMOTZ: You only have properties in the state(`labOrderables` to store `error` and track `loading`, where are the `labOrderables` themselves to be stored after fetching them?

app/js/reducers/labOrders/labOrderableReducer.js
```diff
@@ -0,0 +1,24 @@
...
+} from '../../actions/actionTypes';
+import initialState from '../initialState';
+
+export default (state = initialState.labOrderables, action) => {
+  switch (action.type) {
+    case FETCH_LAB_ORDERABLE_SUCCESS:
+      return {
+        ...state,
+        ...action.orderables,
+      };
```
IAMOTZ: There is a little bit of conflict between the way this reducer is working and the way you have your state structured.
This is what is in the state: 
```
labOrderables: {
    error: false,
    loading: false,
  },
````
And this is what is here in the reducer: 
```
return {
        ...state,
        ...action.orderables,
      };
```
Shouldn't there be a state property to store the `orderables` themselves so that they don't mess up with other data that is in the state(`loading` and `error`)?

File: app/js/actions/actionTypes.js (hunk 2 above)
Comment (topseySuave) [reply app/js/actions/actionTypes.js]: Yes, you are right... @jomadoye

File: tests/reducers/labOrderReducers/labOrderableReducer.test.js (hunk 1 above)
Comment (topseySuave) [reply tests/reducers/labOrderReducers/labOrderableReducer.test.js]: Please more color

File: tests/reducers/labOrderReducers/labOrderableReducer.test.js [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+    const orderable = {
+      labOrderables: [{ id: 1, display: 'outpatient' }],
+    };
+    const expectedState = {
+      labOrderables: [{ id: 1, display: 'outpatient' }],
+      loading: false,
+      error: '',
+    };
+    const actualState = labOrderableReducer(initialState, fetchLabOrderablesSuccess(orderable));
+    expect(actualState).toEqual(expectedState);
```
IAMOTZ: Nice job with adding tests to your implementations 👍 . 
However, one thing to note is that this is a unit test and as a result, you should be testing just this reducer alone, you have to reduce the use of external codes as much as possible. 
I see you are importing the actual actions in this reducer test, what if the actions themselves are not properly written(and they contain bugs), this test would fail and that would make it seem like the reducer has bugs not knowing its the actions.
I suggest you do something like this: 
```
const mockAction = ... // mock the action
const actualState = labOrderableReducer(initialState, mockAction);
```

app/js/actions/actionTypes.js
topseySuave: Done.

app/js/actions/labOrders/labOrderableAction.js
topseySuave: Thanks for the clarification.

app/js/actions/labOrders/labOrderableAction.js
topseySuave: Yes, you are right...

tests/reducers/labOrderReducers/labOrderableReducer.test.js
topseySuave: Done

app/js/reducers/labOrders/labOrderableReducer.js
topseySuave: Done

app/js/reducers/initialState.js
topseySuave: Done, Thanks.

app/js/actions/labOrders/labOrderableAction.js
topseySuave: Done

app/js/actions/labOrders/labOrderableAction.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+/**
+ * @function
+ * @returns {function} - Object consisting lab orderables from API resource
+ * @throws {string} - when an error occurs
+ */
+export const getLabOrderable = () => (dispatch) => {
+  dispatch(loading('FETCH_LAB_ORDERABLES', true));
+  return axiosInstance.get('/concept/da006137-88ca-4d11-ae58-8b4b439afdd6?v=full')
```
mogoodrich: If there isn't already, let's create a top-level constants.js file and move this uuid there and reference by variable here... this is a concept users will need to make sure they have defined in their system so it will be important for all these requirements to be specified in a single file so a user can easily see what they need to set up.

(Long-term this actually should be referenced by a concept mapping, or we should get this from a global property, but I will ticket that in a separate ticket)

topseySuave: No problem @mogoodrich

topseySuave: Changes have been made @mogoodrich

mogoodrich:
Thanks @topseySuave .  We will need to setup some staging data on our test server for this, but I can handle that.

PR #122, opened 2018-07-30
OEUI-181: Fetch Lab Order History

Efosaok:
#### JIRA TICKET NAME
[OEUI-181](https://issues.openmrs.org/browse/OEUI-181) fetch lab order history

#### Summary
- Consume the endpoint for consuming the lab order history of a patient
- Display the data in the past orders accordion


app/js/actions/fetchLabOrders.js
```diff
@@ -0,0 +1,36 @@
...
+  try {
+    const url = uri || `/order?totalCount=true&sort=desc&status=active&order=desc&patient=${patient}&limit=${limit}&t=testorder&v=full`;
+    const response = await axiosInstance.get(url);
+    dispatch(fetchLabOrdersSuccess(response.data));
+  } catch (err) {
+    dispatch(fetchLabOrdersFailure(err.response.data.error.message));
+  }
+};
+
+export default fetchLabOrders;
```
topseySuave: Please have this file moved to the `labOrders` folders.

app/js/components/labOrderEntry/PastOrders.jsx
```diff
@@ -14,9 +15,9 @@ const PastOrders = ({
       </thead>
       <tbody>
         {orders.map(order => (
-          <tr key={order.id}>
-            <td>{order.date}</td>
-            <td>{order.details}</td>
+          <tr key={order.uuid}>
```
topseySuave: @Efosaok In cases like this, that's why the `shortid` package was installed, I would advise you use 
`shortid.generate()` so we don't have similar keys in any scenario.

tests/actions/fetchLabOrderAction.test.js
```diff
@@ -0,0 +1,66 @@
+import fetchLabOrders from '../../app/js/actions/fetchLabOrders';
+
+describe('fetch lab order action test-suite', () => {
+  beforeEach(() => moxios.install());
+  afterEach(() => moxios.uninstall());
+  const limit = 10;
+  const startIndex = 0;
+  const patient = '761d8414-6cb4-4456-be4f-232e6f819e06';
```
topseySuave: A `constants.js` file was created for variables that could be easily changed, please take this off and add to that file.

tests/reducers/fetchLabOrderReducer.test.js
```diff
@@ -0,0 +1,74 @@
...
+          error: true,
+        }
+      }
+    )
+  })
+
+  it('returns initial state if action type is not handled', () => {
+    expect(fetchLabOrderReducer(initialState, {})).toEqual(initialState);
+  })
+})
```
topseySuave: EOL error here @Efosaok

topseySuave: Also, please a folder `labOrders` was created to house all scripts pertaining to lab orders. Please move this file in.

tests/actions/fetchLabOrderAction.test.js
```diff
@@ -0,0 +1,66 @@
...
+        error: 'an error occured',
+      },
+    ];
+
+    const store = mockStore({});
+    await store.dispatch(fetchLabOrders(null, limit, patient));
+    expect(store.getActions()).toEqual(expectedAction);
+    done();
+  })
+});
```
topseySuave: Please move this file to the `labOrderTest` folder

app/js/reducers/fetchLabOrderReducer.js
```diff
@@ -0,0 +1,50 @@
...
+        },
+      };
+    }
+
+    default:
+      return state;
+  }
+};
+
+export default fetchLabOrderReducer;
```
topseySuave: Please move this to `labOrders` folder

app/js/components/labOrderEntry/PastOrders.jsx
Efosaok: from experience, `shortid.generate()` is not usually efficient, it causes bugs if the component re-renders due to change in state, this will work fine as this is the specific id for that particular order that is coming from the database.

tests/actions/fetchLabOrderAction.test.js
Efosaok: When writing tests, I try as much as possible not to use codes outside the `test environment`, I prefer to create a manual mock of every data required for my tests and the `constants.js` file is for use in the actual code, and not for tests.

using the `constants.js` file would also infer that I am not practicing `TDD` as I would be touching a file outside the test environment before writing tests.

tests/reducers/fetchLabOrderReducer.test.js
Efosaok: Thanks for pointing this out

tests/actions/fetchLabOrderAction.test.js
Efosaok: I'll look into this

app/js/reducers/fetchLabOrderReducer.js
Efosaok: I'll look into this

tests/actions/fetchLabOrderAction.test.js
topseySuave: Cool

Efosaok:
@topseysuave, I have moved the files to the appropriate positions

app/js/reducers/fetchLabOrderReducer.js
Efosaok: done

tests/actions/fetchLabOrderAction.test.js
Efosaok: done

tests/reducers/labOrderReducers/fetchLabOrderReducer.test.js
```diff
@@ -0,0 +1,74 @@
...
+          error: true,
+        }
+      }
+    )
+  })
+
+  it('returns initial state if action type is not handled', () => {
+    expect(fetchLabOrderReducer(initialState, {})).toEqual(initialState);
+  })
+})
```
topseySuave: Still have an EOL error here though

tests/reducers/fetchLabOrderReducer.test.js
Efosaok: done

app/js/actions/labOrders/fetchLabOrders.js
```diff
@@ -0,0 +1,36 @@
+import 'babel-polyfill';
+import axiosInstance from '../../config';
```
IAMOTZ: How about you import the babel-polyfill in the entry point of this client codebase so that it can be used in other parts of the codebase without re-importing.

app/js/components/labOrderEntry/LabEntryForm.jsx
```diff
@@ -15,8 +15,11 @@ import {
   removeTestFromDraft,
   addTestToDraft,
   toggleDraftLabOrdersUgency,
+  addDraftLabOrders,
+  deleteDraftLabOrder,
 } from '../../actions/draftLabOrderAction';
```
IAMOTZ: You are not using these two actions you are importing anywhere in this file `addDraftLabOrders`, `deleteDraftLabOrder`, if you do not intend to use it, I suggest you remove it.

app/js/components/labOrderEntry/LabEntryForm.jsx
```diff
@@ -148,43 +165,59 @@ export class LabEntryForm extends PureComponent {
...
           </div>
           <div className="order-form-wrapper">
-            <form className="lab-form simple-form-ui">
-              {this.showFieldSet()}
-            </form>
+            <form className="lab-form simple-form-ui">{this.showFieldSet()}</form>
           </div>
         </div>
         <br />
         {renderActiveOrders()}
```
IAMOTZ: `renderActiveOrders` is still making use of the mock data from `./ordersHistoryMockData`, I am not sure this is supposed to be.

tests/reducers/labOrderReducers/fetchLabOrderReducer.test.js
Efosaok: fixed

app/js/components/labOrderEntry/PastOrders.jsx
```diff
@@ -14,9 +15,9 @@ const PastOrders = ({
...
+            <td>{moment(order.dateActivated).format('DD MMM YYYY')}</td>
+            <td>{order.concept.display}</td>
```
IAMOTZ: I believe `Moment` is a pretty heavy library and if this is all we need it for in this codebase presently: 
```
<td>{moment(order.dateActivated).format('DD MMM YYYY')}</td>
```
I suggest you write a simple javascript utility to do the date formatting.

Efosaok: We would need it later for much more things other than this, so it's better we start using it than write a function now and then it'll become redundant later on.

app/js/components/labOrderEntry/LabEntryForm.jsx
Efosaok: This PR has nothing to do with Active Orders, This PR works with Past orders (history)

app/js/actions/labOrders/fetchLabOrders.js
Efosaok: This has been fixed in another Pull request.

app/js/components/labOrderEntry/LabEntryForm.jsx
Efosaok: fixed

app/js/actions/labOrders/fetchLabOrders.js
IAMOTZ: Owk.

app/js/components/labOrderEntry/PastOrders.jsx
mogoodrich: I think it's definitely makes sense to use a Javascript date library as we will have many occasions to use it across the app.  And we have used moment a lot in the past... however, we are already including date-fns in the application, I believe. Is there a method we could use from there instead?

app/js/components/labOrderEntry/PastOrders.jsx
mogoodrich: I'm not familar with shortid, nor do I understand yet the intriacies of having a key and how React maps to the DOM, but I do agree with @Efosaok, the uuid of the order is guaranteed to be unique, so we wouldn't have to worry about that part of it.

package.json
```diff
@@ -19,6 +20,7 @@
     "deep-freeze": "0.0.1",
     "eslint-config-airbnb-base": "^12.1.0",
     "lodash": "^4.17.5",
+    "moment": "^2.22.2",
```
mogoodrich: I added another comment elsewhere, but can we us date-fns (which we are already including) instead of moment? (Or switch from date-fns to moment)?

tests/actions/fetchLabOrderAction.test.js
mogoodrich: +1 to what @Efosaok said... this is really just mock data that is meaningless outside of the test, not a global constant.

PR #124, opened 2018-07-31
OEUI-201: Use header from react-compnents repo

jomadoye:
**JIRA TICKET NAME**
[OEUI-210](https://issues.openmrs.org/browse/OEUI-201) Use header from react-compnents repo

**Summary**

- This PR gets the Header component from the react component library for reuse


.travis.yml
```diff
@@ -3,8 +3,17 @@ node_js:
 - '8'
 after_success:
 - npm run coveralls
+before_install:
+- cd ..
+- git clone https://github.com/jomadoye/openmrs-react-components.git openmrs-react-components
```
jomadoye: This is temporal the `openmrs-react-components` would soon be published to npm, and this file would be updated to reflect it.

.eslintrc
```diff
@@ -40,6 +40,7 @@
     "no-debugger": 1,
     "no-var": 1,
     "import/no-unresolved": 0,
+    "import/no-extraneous-dependencies": 0,
```
jomadoye: I needed to add this because `npm-link` does not add the package to package.json

.travis.yml
mogoodrich: Thanks for figuring this out @jomadoye.... can we just change this to reference openmrs-react-components in the main OpenMRS project, ie:

https://github.com/openmrs/openmrs-react-components.git

mogoodrich:
hmm, strange that the Travis tests are hanging... do orderentry tests pass when you run locally @jomadoye ?

jomadoye:
Yes @mogoodrich,  they pass locally, but it on Travis it hangs after running some tests and just times out after 10mins.
I'm investigating it

rhenshaw56:
Looks good bro

PR #128, opened 2018-08-01
OEUI-189: Make styling more modular

IAMOTZ:
### JIRA TICKET NAME
[OEUI-189: Make styling more modular](https://issues.openmrs.org/browse/OEUI-189)

### SUMMARY
This Pull Request breaks down the single style sheet that was formerly used in the application. 
It ensures that the styles that are applied to a particular component are modularised to that component such that each styled component has its own style sheet


app/js/components/drugOrderEntry/styles.scss
```diff
@@ -0,0 +1,73 @@
+.drug-order-entry {
```
jomadoye: I think you should add the license at the top of all the files

IAMOTZ: Done! @jomadoye

mogoodrich:
I'm not an expert on css or scss, so defer to others on approving this... :)

app/js/components/drugOrderEntry/styles.scss
```diff
@@ -0,0 +1,82 @@
...
+  .revised-order-text{
+    color:#02473f;
+    font-size: 20px
+  }
+  .flex-row {
+    display: flex;
+    flex-wrap: wrap;
+  }
+  
+  a.disabled {
```
rhenshaw56: you can nest it that way others can apply styles to `a` here...
something like 
```
a {
       .disabled {
            //styles for disabled here
        }
    // other styles for a here
}
```

app/js/components/drugOrderEntry/styles.scss
```diff
@@ -0,0 +1,82 @@
...
+  
+  .small-input {
+    width: 4em;
+  }
+  
+  .medium-input {
+    width: 5em;
+  }  
+
+  .react-paginate-container {
```
rhenshaw56: Nice! 👍

app/js/components/orderEntry/styles.scss
```diff
@@ -0,0 +1,24 @@
...
+    color: red;
+    padding: 0px 8px;
+  }
+  
+  .order-type-button {
+    margin: 10px 5px 10px 0px;
+    outline: none;
+  }
+  
+  .order-type-button.active {
```
rhenshaw56: there is another active class here app/js/components/labOrderEntry/styles.scss, do you think we could somehow combine both ? cos they have the same styles

app/js/components/labOrderEntry/styles.scss
```diff
@@ -0,0 +1,145 @@
...
+    background: #EBEBEB;
+    padding: 10px 10px;
+    margin: 5px 0px;
+    display: flex;
+    align-items: center;
+    position: relative;
+    border-radius: 5px;
+  }
+
+  .draft-list-container .draft-list:hover{
```
rhenshaw56: I suggest you can try to nest this also, although I don't see any way where it would somehow be extended on

app/js/components/patientDashboard/styles.scss
```diff
@@ -0,0 +1,64 @@
...
+    transform: rotate(180deg);
+  }
+  
+  .preformatted-note {
+    margin: 1px;
+    padding: 0px;
+    padding-top: 1px;
+    background: white;
+    color: #e9241d;
+    font-family: "OpenSans", Arial, sans-serif;
```
rhenshaw56: this is been done twice in this file, could you somehow hold it in variable and reuse ?

app/js/components/drugOrderEntry/styles.scss
IAMOTZ: Done! @rhenshaw56

app/js/components/labOrderEntry/styles.scss
IAMOTZ: Done! @rhenshaw56

app/js/components/patientDashboard/styles.scss
IAMOTZ: The `patentDashboard` component has been removed from the codebase consequently, its styles are also removed.

app/js/components/orderEntry/styles.scss
IAMOTZ: Combine the two styles would result in the `active` class to become global, I don't think this would be very nice since it is only used by the two components refrenced and considering the fact that the style itself is very little(just one line), I decided to duplicate it in the two style files.

app/js/components/drugOrderEntry/styles.scss
IAMOTZ: Thanks 😄

PR #131, opened 2018-08-02
OEUI-200: Render Test Panels dynamically

IAMOTZ:
### JIRA TICKET NAME
[OEUI-200: Render the concepts of type LabSet as the panels belonging to a Lab Category](https://issues.openmrs.org/browse/OEUI-200)

### SUMMARY
This Pull Request renders the Test Panels dynamically with the data gotten from the database replacing the mock data used previously.  

**NOTE:** This pull request actually breaks the functionality of adding panels to draft and that is because of the change in the data model. However, this issue would be fixed in this [JIRA TICKET](https://issues.openmrs.org/browse/OEUI-212)

### DIFF
--- a/app/js/components/labOrderEntry/LabPanelFieldSet.jsx
+++ b/app/js/components/labOrderEntry/LabPanelFieldSet.jsx
@@ -1,21 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { panelData } from './labData';
 import '../../../css/grid.scss';
 
-const LabPanelFieldSet = ({ selectedPanelIds, handleTestSelection }) => (
+
+const formatPanelName = (panelName) => {
+  const name = panelName;
+  return name.replace(/panel/i, '').trim();
+};
+
+const LabPanelFieldSet = ({ selectedPanelIds, handleTestSelection, panels }) => (
   <fieldset className="fieldset">
     <legend>Panels</legend>
-    {panelData.map(panel => (
-      <button
-        id="panel-button"
-        className={selectedPanelIds.includes(panel.id) ? 'active lab-tests-btn' : 'lab-tests-btn'}
-        type="button"
-        key={`${panel.id}`}
-        onClick={() => handleTestSelection(panel, 'panel')}>
-        {panel.name}
-      </button>
-    ))}
+    <div className="panel-box">
+      {
+        panels.map(panel => (
+          <button
+            id="panel-button"
+            className={(selectedPanelIds.includes(panel.uuid)) ? 'active lab-tests-btn' : 'lab-tests-btn'}
+            type="button"
+            key={`${panel.uuid}`}
+            onClick={() => handleTestSelection(panel, 'panel')}
+          >
+            {formatPanelName(panel.display.toLowerCase())}
+          </button>
+        ))
+      }
+    </div>
   </fieldset>
 );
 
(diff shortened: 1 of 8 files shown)


app/js/components/labOrderEntry/LabPanelFieldSet.jsx
```diff
@@ -1,27 +1,39 @@
...
-import { panelData } from './labData';
 import '../../../css/grid.scss';
 
+
+const formatPanelName = (panelName) => {
+  const name = panelName;
+  const formattedName = name.split(' ')
+    .map(splittedName => splittedName[0].toUpperCase() + splittedName.substr(1).toLowerCase())
+    .join(' ')
+    .replace(/panel/i, '')
```
topseySuave: where did the `i` come from?

IAMOTZ: The `i` is a regular expression flag.
It is used to ignore case-sensitivity so that strings like `Panel`, `PaNel`, `paneL` e.t.c would all be a match.
You can find more details about regular expression flags [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Switching_words_in_a_string).

topseySuave: I thought it was a path

tests/components/labOrderEntry/LabEntryForm.test.jsx
```diff
@@ -173,6 +175,13 @@ describe('Component: LabEntryForm', () => {
     expect(dispatch).toBeCalled();
   });
 
+  it('should toggle the urgency state of a test', () => {
+    const instance = getComponent().instance();
+    const dispatch = jest.spyOn(props, 'dispatch');
+    instance.handleUrgencyChange();
+    expect(dispatch).toBeCalled();
+  }) 
+
```
Efosaok: is there a way you could simulate and assert based on the *DOM* itself instead of calling the component methods directly? this is a good situation to use mount if the dom element that would trigger the `handleUrgencyChange` method is in a child component.

the test-case here doesn't seem to flow with its message.

please refer to [this](https://medium.freecodecamp.org/the-right-way-to-test-react-components-548a4736ab22) when you have time.

IAMOTZ: This method is actually not used in this component(`LabEntryForm`), it is passed down to another component `LabDraftOrder` as a prop. I believe the testing based on the result of the DOM should be done in the test for that component and not this.
This is all the method contains in this component:
```
handleUrgencyChange = (order) => {
    this.props.dispatch(toggleDraftLabOrdersUgency(order));
 }
```
Consequently, testing for if `dispatch` is called should be enough assertion.

app/js/components/labOrderEntry/LabPanelFieldSet.jsx
rhenshaw56: @IAMOTZ can't you still use the `text-transform` property here also ?

IAMOTZ: Yes, I can.
I have updated the function.
Thanks, @rhenshaw56

tests/components/labOrderEntry/LabEntryForm.test.jsx
Efosaok: ok

PR #141, opened 2018-08-08
OEUI-219: Make Only Top level Lab-Tests Orderable

IAMOTZ:
### JIRA TICKET NAME
[OEUI-219: Only tests that are top-level should be orderable.](https://issues.openmrs.org/browse/OEUI-219)

### SUMMARY
In the previous implementation of panel/test selection when making Lab Order, any test in the tests-box can be selected regardless of if it is a standalone test or not. This pull request removes all the test that are not standalone from the tests-box consequently making only standalone tests orderable.


mogoodrich:
Looks good to me!

PR #142, opened 2018-08-08
OEUI-187: Rename action suffix types

rhenshaw56:
### JIRA TICKET NAME
[OEUI-187](https://issues.openmrs.org/browse/OEUI-187) Rename action suffix types


app/js/actions/labOrders/fetchLabOrders.js
```diff
@@ -7,30 +7,15 @@ import {
...
 
-const fetchLabOrdersLoading = status => ({
-  status,
-  type: FETCH_LAB_ORDERS_LOADING,
+const fetchLabOrders = (
+  uri,
+  limit = 5,
+  patient,
+) => ({
+  type: 'FETCH_LAB_ORDERS',
```
rhenshaw56: We don't need to handle `SUCCESS`, `FAILURE` or `LOADING` cos the middleware does that for us. We just return a thunk as an async action and since it returns a promise, it'll map directly to whatever value it returns ie. pending, resolved or rejected

IAMOTZ: Good job @rhenshaw56 
This is such a nice refactoring. With `redux-promise-middleware`, we can reduce the amount of codes we write and it would even make the codebase better to understand.

But there is something I am still confused about with the explanation you gave above: 
My understading on how to use the `redux-promise-middleware` is to return an action(as an object), but the action can have a promise payload. The explanation above states that _`...We just return a thunk... `_, I am not sure of what you are referring to with this since you are not even returning a `thunk` in the code you wrote. Can you please provide clarity on that?

IAMOTZ: I see that you only made this refactoring(using `redux-promise-middleware` pattern) in `fetchLabOrdersAction.js` and `labConceptsAction.js`, do you plan on refactoring other action files to also follow this new pattern?

rhenshaw56: yes @IAMOTZ I plan to do that! currently doing that but for each file I touch, the actions and reducers tests break, code breaks in the UI also based on maybe how it was imported and how the reducer read the data, so it's gonna be a bumpy ride.

rhenshaw56: @mogoodrich yes it will generate those actions automatically depending on the state of the promise (pending, fulfilled, failed), using sagas combined with redux-promise-middleware is on it's own very powerful in the redux arsenal, cos with sagas, it's easier to trigger optimistic updates and the promise middleware makes it easier to reduce logic for fetching data which means easier debugging and reduction in code size, reducing the number of problems one could have lol! yeah, we could add it to be used alongside the sagas

app/js/middlewares/responseHandlerMiddleware.js
```diff
@@ -0,0 +1,69 @@
...
+      return next({
+        ...action,
+        type: action.type,
+        data: action.payload.data,
+        meta: action.meta || {},
+      });
+    }
+    return next(action);
+  };
+}
```
mogoodrich: @rhenshaw56  I'm still learning how all this works... can you explain what the theory is behind this middleware?  That is, how do we determine what logic goes here as opposed to in a reducer?  Am I right that this seems to handle the kind of stuff that would go in a saga if we were using sagas in this project?

rhenshaw56: @mogoodrich not really! so in sagas you're watching for events to dispatch actions which is different in this case. so what happens is;
```
const middlewares = [
  thunk,
  sagaMiddleware,
  errorMiddleware,
  promiseMiddleware({ promiseTypeSuffixes }),
  responseHandlerMiddleware,
];
```
firstly, the `thunk` middleware is declared first so it tries to handle everything that is a `thunk` ie any action that is coming as a function, resolves it and passes it on to the next middleware, but the saga middleware can't handle that because thunk will return an object instead the `saga` middleware watches for sagas and executes them and passes it on the next middleware which is the errorMiddleware`, here we're just trying to catch errors resulting from failed promises and then passing them on to the next middleware which is the `promiseMiddleware` that handles pending promises and resolves unrejected ones. the `responseHandler` middleware is just for formatting responses from the server and the reason why it's placed after the `promiseMiddleware` is that it only handles those being passed down by the `promiseMiddleware` after thy've been resolved cos they come in as promises. so I noticed that in some actions we were returning `response.data`, `response.data.results`, and in most cases, we'd have to format it in some way before it gets to the reducer.

mogoodrich: Got is, so the "responseHandlerMiddleware" is generally responsible for making sure that the async responses get formatting into some sort of standard format?  When I'm adding a new feature that fetching and manipulating data, just trying to figure out how I determine what logic should go in the responseHandlerMiddleware and what should go in a reducer.

@rhenshaw56

rhenshaw56: @mogoodrich yeah, that's right and if you check line 59 of the responsHandlerMiddleware, every request made henceforth will by default be gotten as action.data in the every reducer

mogoodrich: Sounds good!

PR #145, opened 2018-08-10
OEUI-222: Remove the inpatient/outpatient option for drug orders

IAMOTZ:
### JIRA TICKET NAME
[OEUI-222: Take out the inpatient and outpatient care settings](https://issues.openmrs.org/browse/OEUI-222)

### SUMMARY
This pull request removes the inpatient/outpatient tabs(and settings) that are on the Drug Order page in the application. Consequently, the drug order functionality treats the patients as outpatients.

### DIFF
--- a/app/js/components/drugOrderEntry/addForm/AddForm.jsx
+++ b/app/js/components/drugOrderEntry/addForm/AddForm.jsx
@@ -49,9 +49,6 @@ export class AddForm extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.careSetting.display !== this.props.careSetting.display) {
-      this.clearDrugForms();
-    }
     return (
       Object.keys(this.props.draftOrder).length ||
       Object.keys(this.props.editOrder).length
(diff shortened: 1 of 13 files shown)


app/js/components/drugOrderEntry/addForm/AddForm.jsx
```diff
@@ -149,7 +146,7 @@ export class AddForm extends React.Component {
         item = drugDispensingUnits.filter(unit => unit.display === value).length > 0;
         break;
       default:
-        item = value.length > 0;
+        // does nothing
```
topseySuave: Can you elaborate and through more colour.

IAMOTZ:
Hey @Efosaok 
_` ...I noticed you just took out the UI components related to care setting...`_: I did not take out the UI components related to `care setting`, I took out the UI components related to `Outpatient Care Setting`.

The `care setting` action creators and reducers fetch both the `Inpatient Care Setting` and the `Outpatient Care Setting`, they are still very much needed since `Outpatient Care Setting` is what is now used for creating Drug Ordres.

app/js/components/drugOrderEntry/addForm/AddForm.jsx
IAMOTZ: That particular line of code `item = value.length > 0;` does nothing, there is actually no switch value that would result in that default. You can confirm that form the way the `handleValidation` method is used in `FreeText.jsx` and `StandardDose.jsx`

The reason I removed it is because it is dropping the code coverage of the project.

mogoodrich:
I haven't been following this closely and haven't looked at the action creators, but I think I agreed with @IAMOTZ ... it's likely that "Inpatient" care setting will be reintroduced at some point, so if the action creators are generic creators used for fetching care setting, it would seem to make sense to keep them around and just add a comment to them "not currently used" for the ones that aren't currently used.

PR #148, opened 2018-08-14
OEUI-217: Fetch orders dynamically on the homepage

jomadoye:
**JIRA TICKET NAME**
[OEUI-217](https://issues.openmrs.org/browse/OEUI-217) Generate homepage data dynamically

**Summary**

- This PR gets the orders dynamically


app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -1,30 +1,62 @@
...
-      </Accordion>
-    ))}
-  </React.Fragment>
-);
+export class OrdersTable extends PureComponent {
+  componentDidMount() {
+    this.props.dispatch(fetchOrder(null, 5, this.props.patient.uuid));
+  }
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (this.props.patient.uuid !== prevProps.patient.uuid) {
```
Efosaok: you do not seem to use the `prevState` and `snapshot` variables here, please take it out.

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -1,30 +1,62 @@
...
-          />
-        ) : (
-          <LabOrderDetails urgency={order.urgency} orderer={order.orderer} />
-        )}
-      </Accordion>
-    ))}
-  </React.Fragment>
-);
+export class OrdersTable extends PureComponent {
+  componentDidMount() {
```
Efosaok: I am aware that there is a parent component for this, why not move the `componentLifeCycle` methods into that Parent Component.

app/js/actions/fetchOrder.js
```diff
@@ -0,0 +1,16 @@
+import 'babel-polyfill';
+import axiosInstance from '../config';
+
+import { FETCH_ORDERS } from './actionTypes';
+
+const fetchOrders = (
+  uri,
```
Efosaok: The name of this file is not the same as its `export-default`, `fetchOrder` vs `fetchOrders`.

tests/actions/fetchOrdersAction.test.js
```diff
@@ -0,0 +1,57 @@
+import fetchOrders from '../../app/js/actions/fetchOrder';
+
+import {
+  FETCH_ORDERS_LOADING,
+  FETCH_ORDERS_SUCCESS
+} from '../../app/js/actions/actionTypes';
+describe('fetchOrders action', () => {
```
Efosaok: since this is a test, you do not need to import real instances to use in the test, it is always advisable to mock if possible, it'd be better if you just used the string-literals here directly instead of importing your code.

tests/actions/fetchOrdersAction.test.js
```diff
@@ -0,0 +1,57 @@
...
+        status: 201,
+        response: [{}],
+      },
+    );
+    const expectedTypes = [
+      FETCH_ORDERS_LOADING,
+      FETCH_ORDERS_SUCCESS,
+    ];
+    const store = mockStore({});
+    return store.dispatch(fetchOrders(null, limit, patient)).then(() => {
```
Efosaok: since the `await` syntax is already available in this function, why still using `.then` to resolve `promises` in this function .

tests/actions/fetchOrdersAction.test.js
```diff
@@ -0,0 +1,57 @@
...
+        response: [{}],
+      },
+    );
+    const expectedTypes = [
+      FETCH_ORDERS_LOADING,
+      FETCH_ORDERS_SUCCESS,
+    ];
+    const store = mockStore({});
+    return store.dispatch(fetchOrders(uri, null, patient)).then(() => {
+      const dispatchedActions = store.getActions();
```
Efosaok: refer to above comment

tests/actions/fetchOrdersAction.test.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+    );
+    const expectedTypes = [
+      FETCH_ORDERS_LOADING,
+      FETCH_ORDERS_SUCCESS,
+    ];
+    const store = mockStore({});
+    return store.dispatch(fetchOrders(null, limit, patient)).then(() => {
+      const dispatchedActions = store.getActions();
+      const actionTypes = dispatchedActions.map(action => action.type);
```
Efosaok: why `mapping` here.

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -0,0 +1,50 @@
...
+    loading: false,
+  },
+};
+
+const fetchOrdersReducer = (state = initialState, action) => {
+  switch (action.type) {
+    case 'FETCH_ORDERS_SUCCESS':
+      return {
+        ...state,
+        orders: action.payload.data,
```
rhenshaw56: I don't think it would be accessible as `action.payload.data` if it's a `SUCCESS` cos the responseMiddleware will format it as `action.data`

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -0,0 +1,50 @@
...
+          ...state.status,
+          fetched: true,
+          loading: false,
+        },
+      };
+
+    case 'FETCH_ORDERS_FAILURE':
+      return {
+        ...state,
+        errorMessage: action.error,
```
rhenshaw56: now here would be `action.payload`

app/js/components/orderEntry/OrdersTable.jsx
jomadoye: The `parent`, `grandParent` and `great-grandparents` of this component are all `presentational` components. 
If we decide to make the parent a `container` component, then we should also make it's grand-parent` a container component and pass down the data as a prop. 
I believe it's better to have one single container component down the line, than passing down props up to 3 layers to make it  `presentational`.
Also remember each component would be re-rendered, each time it's props is updated when it is being passed down.

tests/actions/fetchOrdersAction.test.js
jomadoye: I believe the main purpose of declaring action types as constants and exporting them, is because of the need to mitigate human typographical error. Importing a const isn't necessarily mocking the actions.
it's just to ensure that no typos are being made when testing the actions.

app/js/reducers/fetchOrdersReducer.js
jomadoye: On the contrary, it works fine as is, also my test breaks when I implement your changes and returns undefined.
But it works fine as is, please help me investigate if you still feel, that should not be the expected result.

app/js/reducers/fetchOrdersReducer.js
jomadoye: Ditto!

tests/actions/fetchOrdersAction.test.js
jomadoye: regarding importing real instances.
we import real `components`, `actions` and `reducers` to test them, I don't see anything wrong in importing real `actionTypes`.

app/js/reducers/fetchOrdersReducer.js
rhenshaw56: yeah, so it will work fine cos the payload would still be in action anyways but `action.data` is still the same thing as `action.payload.data`, for your test breaking, you would need to mock your response as `action.data` and get rid of the payload bit.

try logging action in the reducers and see what gives.

jomadoye:
All comments have been Fixed/Replied

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -1,30 +1,62 @@
...
+    this.props.dispatch(fetchOrder(null, 5, this.props.patient.uuid));
+  }
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (this.props.patient.uuid !== prevProps.patient.uuid) {
+      this.props.dispatch(fetchOrder(null, 5, this.props.patient.uuid));
```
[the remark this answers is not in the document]
Efosaok: I think it does, just for the sake of keeping the parameters in place, That's why it's supposed to be the last parameter or he could omit it and explicitly declare what `this.props.patient.uuid` represents..

tests/actions/fetchOrdersAction.test.js
jomadoye: Fixed, thanks.

tests/actions/fetchOrdersAction.test.js
jomadoye: Fixed, thanks.

app/js/actions/fetchOrder.js
```diff
@@ -0,0 +1,16 @@
...
+  limit = 5,
```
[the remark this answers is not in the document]
jomadoye: Fixed, thanks.

app/js/components/orderEntry/OrdersTable.jsx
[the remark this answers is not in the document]
jomadoye: Fixed, thanks.

app/js/reducers/fetchOrdersReducer.js
jomadoye: Fixed, thanks.

app/js/reducers/fetchOrdersReducer.js
jomadoye: Fixed, thanks.

app/js/components/orderEntry/OrdersTable.jsx
mogoodrich: I think I agree with @jomadoye ... also, we'd want this OrdersTable component to be able to be used in multiple contexts, within different presentations/layouts, so doesn't it make sense for it to be responsible for fetching the data it plans to display? Or is there a different paradigm we should be following?

PR #152, opened 2018-08-15
OEUI-220: Implement Sorting And Filtering

Efosaok:
#### JIRA TICKET NAME
[OEUI-220](https://issues.openmrs.org/browse/OEUI-220)

#### Summary
- Implement sorting and Filtering


mogoodrich:
LGTM once the tests are fixed.

app/js/components/orderEntry/SortAndFilter.jsx
```diff
@@ -0,0 +1,37 @@
...
+      </select>
+    </div>
+    <div className="sort-status">
+      <label className="label-sort">Status:</label>
+      <select className="sort-select">
+        <option>All</option>
+        <option>Active</option>
+        <option>Inactive</option>
+      </select>
+    </div>
```
IAMOTZ: Please add some comments(inline or block) as to why 
```
<div className="sort-status">
  ....
</div>
```
is hidden in the view.

tests/components/orderentry/SortAndFilter.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+}
+const wrapper = shallow(<SortAndFilter store={store} />).dive();
+
+describe('sort and Filter component test-suite', () => {
+  it('renders properly', () => {
+    const selectElement = wrapper.find('select').at(0)
+    selectElement.simulate('change', { target: { value: 'testorder' }});
+    expect(wrapper.exists()).toBeTruthy();
+    expect(selectElement.exists()).toBeTruthy();
+  });
```
IAMOTZ: I don't think this assertion: 
```
expect(selectElement.exists()).toBeTruthy();
```
is enough to test that the component `SortAndFilter` is rendering properly.
I suggest you generate a snapshot to confirm that the component is properly rendered.

tests/components/orderentry/SortAndFilter.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+});
```
IAMOTZ: After simulating change with:
```
selectElement.simulate('change', { target: { value: 'testorder' }});
```
the component is supposed to call a function with some arguments, please add test for that.

tests/reducers/fetchOrdersReducer.test.js
```diff
@@ -17,65 +18,144 @@ describe('fetchorders reducer test-suite', () => {
...
+      {
+        type: 'SORT_AND_FILTER',
+        sortBy: 'type',
+        value: 'drugorder',
+      },
+    )).toEqual({
+      ...initialState,
+      orders: {
+        results: [],
+      },
```
IAMOTZ: This test suit states
```
it('uses the data in localStorage if the filter array is empty'...
```
But I don't seem to find anywhere in this test suit that the `localStorage` is used, am I missing something?

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -42,6 +58,16 @@ const fetchOrdersReducer = (state = initialState, action) => {
       };
     }
 
+    case 'SORT_AND_FILTER': {
+      return {
+        ...state,
+        orders: {
+          ...state.orders,
+          results: filter(state.orders.results, action.sortBy, action.value),
+        },
```
IAMOTZ: I don't think mutating the actual result from the database is nice. We shouldn't mutate the redux state directly like this. One edge case to note is if there are other functionalities in the application making use of this result, they would start breaking as a result of this change.
I have two suggestion to improve this block of code:
- keep the result of the filter somewhere else.
- use the state of the component holding the orders-data to track the filtering

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -8,9 +10,23 @@ const initialState = {
...
 };
 
+const filter = (from, sortBy, value) => {
+  const savedOrders = JSON.parse(localStorage.getItem('ORDERS')) || from;
+  switch (value) {
+    case 'all':
+      return savedOrders;
+    default:
+      if (from.length === 0) {
+        return savedOrders.filter(order => order[sortBy] === value);
```
IAMOTZ: Some concerns on this line of code 
```
 if (from.length === 0) {
  ....
}
```
Let's assume this is the initial result:
```
[{ ... type: 'drugorder' ... }, { ... type: 'drugorder' ... }, { ... type: 'testorder' ... }, { ... type: 'xxxorder' ... }, ]
```
After filtering out all the `drugorder`, this is what `from` would be like
```
[{ ... type: 'testorder' ... }, { ... type: 'xxxorder' ... }, ]
```
And then I filter out all the `xxxorder`, `from` would become:
```
[{ ... type: 'testorder' ... }]
```
Lets say I want to get back all the `drugorders`, do you think I would be able to do that? don't you think this is breaking the functionality?

app/js/reducers/fetchOrdersReducer.js
Efosaok: The logic for state change in a redux-application is done in the reducers, therefore this is appropriate, and it is generally a bad practice to fetch data from the redux state and then store the data in your components state again and then manipulate data inside the component when the reducers should have acted on it in the first place.

The actual result is manipulated by the reducer but not lost, it is stored in the local storage and the coincidence that this would break the application is next to none as the homepage is currently the only component using this reducer and the result's array always has an initial state of an empty array. therefore, during implementation, any component that would feed from the results array should be made to work with at least the *initial state* and the current implementation makes sure the result array always have at least the value at the *initial-state([])*.

app/js/reducers/fetchOrdersReducer.js
Efosaok: hmm, I'll address this

app/js/reducers/fetchOrdersReducer.js
Efosaok: I get your point at @mogoodrich, but we can't apply any filter at that point cus the `.map` is meant to display the data that is in the `results` array and the filter is applied by the user once the DOM is loaded at least once.

If you aren't comfortable with introducing localStorage and manipulating data in the reducers, I'll be moving the logic to the component state now .

tests/components/orderentry/SortAndFilter.test.jsx
Efosaok: addressed

tests/components/orderentry/SortAndFilter.test.jsx
Efosaok: addressed

tests/reducers/fetchOrdersReducer.test.js
Efosaok: addressed

Efosaok:
I have attended to all issues @IAMOTZ @mogoodrich 
I believe this is ready to go in

app/js/actions/fetchOrders.js~HEAD
```diff
@@ -0,0 +1,15 @@
...
+  uri,
+  patient,
+  limit = 5,
+) => ({
+  type: FETCH_ORDERS,
+  payload: axiosInstance
+    .get(uri || `/order?totalCount=true&sort=desc&status=active&patient=${patient}&limit=${limit}&v=full`),
+});
+
+export default fetchOrders;
```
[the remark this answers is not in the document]
Efosaok: addressed

app/js/components/orderEntry/SortAndFilter.jsx
Efosaok: addressed

app/js/actions/fetchOrders.js~HEAD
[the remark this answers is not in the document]
IAMOTZ: You can just remove this file since it is not needed.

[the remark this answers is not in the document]
Efosaok: it is needed.

jest.config.js
```diff
@@ -16,7 +15,8 @@ module.exports = {
   testEnvironment: "jsdom",
   roots: ['<rootDir>'],
   setupFiles: [
-    '<rootDir>/tests/setup.js'
+    '<rootDir>/tests/setup.js',
+    'jest-localstorage-mock',
```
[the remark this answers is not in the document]
Efosaok: removed

app/js/reducers/fetchOrdersReducer.js
Efosaok: That line is necessary. with your current code, you are neglecting the fact that a user might filter from the pre-filtered results as well and not from the whole results all the time, a perfect instance of that is when there would be more than one sort query to sortBy(e.g status, dateActivated, e.t.c). my current implementation goes beyond what we have now and accomodates future use-cases.

let me give you an instance;
let's say the data returned from the db is;

`[{ id: 1, type: 'drugorder', xxxquery: 'active' }, { id: 2, type: 'drugorder', xxxquery: inactive }, { id: 3, type: 'testorder', xxxquery: 'active'}, ]`

and a user queries by type with a value of drugorder. the filtered results would be;
`[{ id: 1, type: 'drugorder', xxxquery: 'active' }, { id: 2, type: 'drugorder', xxxquery: inactive } ]`

if a user then decide to sort using the alternate query,maybe, based on what is available would want to see only drug orders that are active, your code would filter from the first original results. thus it'll produce

`[{ id: 1, type: 'drugorder', xxxquery: 'active' }, { id: 3, type: 'testorder', xxxquery: 'active'}, ]`

and that is an undesired result.
my current implementatiion would still filter from the already filtered result and produce;

`[{ id: 1, type: 'drugorder', xxxquery: 'active' } ]` .

I hope you get the point..

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -1,5 +1,10 @@
+import { all } from 'redux-saga/effects';
+
 const initialState = {
-  orders: {},
+  orders: {
+    results: [],
```
mogoodrich: I know this was the way it was previously, but this might be good time to refactor to just:

`orders: []`

... and then update the components that map in this state to the new format.

Having the orders nested under "results" seems unnecessary, and is now inconsistent with the new:

`filteredOrders: []
`

app/js/reducers/fetchOrdersReducer.js
IAMOTZ: Nice catch @Efosaok

app/js/actions/fetchOrders.js~HEAD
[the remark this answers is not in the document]
Efosaok: yes, did not notice earlier..

app/js/components/orderEntry/Accordion.jsx
```diff
@@ -24,7 +26,7 @@ class Accordion extends React.Component {
                 <i className="toggle-icon icon-caret-right small rotate90" />
               )}
             </span>
-            27 jun 2018
+            {format(date, 'MM/DD/YYYY')}
```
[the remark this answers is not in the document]
Efosaok: addressed

app/js/components/labOrderEntry/LabPanelFieldSet.jsx
```diff
@@ -11,8 +11,8 @@ const formatPanelName = (panelName) => {
 
 const formatToolTipData = (setMembers) => {
   const testArray = [];
-  setMembers.map(test => testArray.push(test.display));
-  return testArray;
+  return setMembers.map(test => test.display);
+  // return testArray;
```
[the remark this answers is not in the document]
Efosaok: addressed

app/js/reducers/fetchOrdersReducer.js
Efosaok: addressed in comments below

PR #154, opened 2018-08-16
OEUI-228 : displaying proper action buttons for the different order types on the order home page

rhenshaw56:
**JIRA TICKET**

[OEUI-228](https://issues.openmrs.org/browse/OEUI-228) - Display proper action buttons for the different order types on the order home page

### DIFF
--- a/app/js/components/orderEntry/OrderHeader.jsx
+++ b/app/js/components/orderEntry/OrderHeader.jsx
@@ -1,13 +1,22 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
+import IconButton from '../button/IconButton';
 
 const OrderHeader = ({ status, orderable }) => (
   <Fragment>
     <th>{orderable}</th>
     <th>{status}</th>
-    <th>
-      <i className="icon-edit" />
-      <i className="icon-delete" />
+    <th className="order-actions-btn">
+      <IconButton
+        iconClass="icon-pencil"
+        iconTitle="EDIT"
+        onClick={() => {}}
+      />
+      <IconButton
+        iconClass="icon-remove"
+        iconTitle="DISCONTINUE"
+        onClick={() => {}}
+      />
     </th>
   </Fragment>
 );
--- a/app/js/components/orderEntry/styles.scss
+++ b/app/js/components/orderEntry/styles.scss
@@ -171,3 +171,21 @@
     max-height: 500px;
   }
 }
+
+.order-actions-btn {
+  display: flex;
+  span {
+    width:40%;
+    a {
+      i.scale {
+        text-decoration: none;
+         color: black;
+      }
+    }
+    :hover{
+      transform: scale(1.5);
+      -webkit-transform: scale(1.5);
+      -moz-transform: scale(1.5);
+    }
+  }
+}
--- a/tests/components/button/IconButton.test.jsx
+++ b/tests/components/button/IconButton.test.jsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import IconButton from '../../../app/js/components/button/IconButton';
+
+let props;
+let mountedComponent;
+
+const getComponent = () => {
+  if (!mountedComponent) {
+    mountedComponent = shallow(<IconButton {...props} />);
+  }
+  return mountedComponent;
+};
+
+describe('Component: IconButton', () => {
+  beforeEach(() => {
+    props = {
+      iconClass: 'icon-remove',
+      iconTitle: 'Header',
+      onClick: jest.fn(),
+    };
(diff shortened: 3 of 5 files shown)


IAMOTZ:
Hi @rhenshaw56
I think you want to double-check the Pull Request title, `Display action buttons for draft items`.
The ticket referenced in the description is about `displaying proper action buttons for the different order types on the order home page`, I think there is a wording error somewhere in between.

app/js/components/orderEntry/OrderHeader.jsx
```diff
@@ -1,13 +1,22 @@
...
+    <th className="draft-action-btns">
+      <IconButton
+        iconClass="icon-pencil"
+        iconTitle="EDIT"
+        onClick={() => {}}
+      />
+      <IconButton
+        iconClass="icon-remove"
+        iconTitle="DISCARD"
+        onClick={() => {}}
```
IAMOTZ: `iconTitle="DISCARD"`, I think this is supposed to be `iconTitle="DISCONTINUE"`. Take a look at this snapshot from the previous implementation:
screen shot 2018-08-17 at 11 37 42 am

app/js/components/orderEntry/OrderHeader.jsx
```diff
@@ -1,13 +1,22 @@
...
+      />
```
IAMOTZ: From the component composition above, I see you rendered edit/discontinue icon for both lab-orders and drug-orders. I don't think this is supposed to be, the icon rendering has to be dynamic, edit/discontinue-icon for drug orders while urgency-icon for lab orders.

tests/components/button/IconButton.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    mountedComponent = undefined;
+  });
+
+  it('renders properly', () => {
+    const component = getComponent();
+    expect(component).toMatchSnapshot();
+  });
+
+  it('has a hndler for click events', () => {
+    const event = {
```
IAMOTZ: Typo error here: _`... hndler...`_

app/js/components/orderEntry/OrderHeader.jsx
```diff
@@ -1,13 +1,22 @@
...
 
 const OrderHeader = ({ status, orderable }) => (
   <Fragment>
     <th>{orderable}</th>
     <th>{status}</th>
-    <th>
-      <i className="icon-edit" />
-      <i className="icon-delete" />
+    <th className="draft-action-btns">
+      <IconButton
```
IAMOTZ: `className="draft-action-btns"`: This class name makes it seem like the action buttons in this component are in a draft container but that is not the case, so consider refactoring the class naming.

app/js/components/orderEntry/styles.scss
```diff
@@ -171,3 +171,21 @@
...
+         color: black;
+      }
+    }
+    :hover{
+      transform: scale(1.5);
+      -webkit-transform: scale(1.5);
+      -moz-transform: scale(1.5);
+    }
+  }
+}
```
IAMOTZ: Some concerns on this class name`draft-action-btns`.
The styles inside the class is actually affecting some elements on the orders home page, I understand that the same styles may also be needed for the draft, but you should consider using a more generic name that would reflect the fact that the class is not for the draft action buttons alone.

rhenshaw56: @IAMOTZ I don't see how it would affect anything else anywhere since the stylesheet itself is scoped to `OrderEntryPage.jsx` and yes the styles are for the buttons on the home page ? am I missing something

app/js/components/orderEntry/OrderHeader.jsx
rhenshaw56: @IAMOTZ the implementation on this ticket followed the designs from the mockup and it wasn't stated on the ticket either, but after further clarification from David, he said to make the lab-orders non-editable for now, so not even urgency toggling for now which I have fixed on OEUI-225 which is a sequel to this.

IAMOTZ: Okay.
Thanks for the clarification. 👍

app/js/components/orderEntry/styles.scss
IAMOTZ: Hey @rhenshaw56 
I was not talking about the effect of the styles, the comment was only about the name of the class holding the styles `draft-action-btns`. Since the styles are both used on the `Draft` and the `Orders` view, I was suggesting you use a more generic name like  `order-action-btn`.
Do you get?

rhenshaw56: done.

PR #168, opened 2018-08-29
OEUI-227: Toggle Urgency of lab orders from home page list

Efosaok:
### **JIRA TICKET NAME**
[OEUI-227](https://issues.openmrs.org/browse/OEUI-227)

### **Summary**
- implement toggling urgency from homepage orders list

- [x] I can successfully toggle urgency of lab orders


tests/components/orderentry/OrdersTable.test.jsx
```diff
@@ -76,50 +110,145 @@ describe('Orders component test-suite', () => {
...
+            ],
+          },
+          status: {
+            added: true,
+          },
+        },
+      });
+      expect(props.dispatch.mock.calls[props.dispatch.mock.calls.length - 1][0].type)
+        .toEqual('TOGGLE_URGENCY');
+      expect(global.toastrMessage).toEqual('urgency successfully changed')
```
dondrzzy: Missing semicolon.

tests/reducers/fetchOrdersReducer.test.js
```diff
@@ -214,6 +214,64 @@ describe('fetchorders reducer test-suite', () => {
...
+      },
+      filteredOrders: [
+        {
+          uuid: 1,
+          type: 'drugorder',
+          urgency: 'STAT',
+        },
+      ],
+    });
+  })
```
dondrzzy: Here too. Otherwise it looks good.

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -9,21 +9,86 @@ import DrugOrderDetails from './DrugOrderDetails';
...
+    const { previousOrderId, recentUrgency } = this.state;
+    const {
+      status: { added, error },
+      errorMessage,
+      labOrderData,
+    } = this.props.createLabOrderReducer;
     if (this.props.patient.uuid !== prevProps.patient.uuid) {
       this.props.dispatch(fetchOrders(null, this.props.patient.uuid));
     }
+    if (added && labOrderData !== prevProps.createLabOrderReducer.labOrderData) {
```
rhenshaw56: please what is does the variable name `added` represent ?

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -9,21 +9,86 @@ import DrugOrderDetails from './DrugOrderDetails';
...
-    this.props.dispatch(setSelectedOrder({ order: { ...order, status: 'EDIT' }, currentOrderType: DRUG_ORDER, activity: 'EDIT' }));
-  }
+    this.props.dispatch(setSelectedOrder({
+      order: { ...order, status: 'EDIT' },
+      currentOrderType: DRUG_ORDER,
+      activity: 'EDIT',
+    }));
+  };
+
+  toggleUrgency = (order) => {
```
rhenshaw56: so looking at this file, there are different `togglUrgency`'s already. one is an action and the other is a handler for actually toggling urgency, do you mind renaming this to `handleUrgencyToggling` so it follows the convention of an event handler for clicks ?

app/js/reducers/fetchOrdersReducer.js
```diff
@@ -24,6 +24,13 @@ const filter = (allResults, filteredResults, sortBy, value) => {
   return newFilter;
 };
 
+const toggleUrgency = (previousOrderId, newOrderId, urgency, orders) => {
+  const toggledUrgencyOrders = orders.map(order =>
```
rhenshaw56: this variable name doesn't read so good... `toggledUrgencyOrders ` ?

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -9,21 +9,86 @@ import DrugOrderDetails from './DrugOrderDetails';
...
     }
+    if (added && labOrderData !== prevProps.createLabOrderReducer.labOrderData) {
+      this.props.dispatch(toggleUrgency(
+        previousOrderId,
+        labOrderData.orders[0].uuid,
+        recentUrgency,
+      ));
+      successToast('urgency successfully changed');
+    }
+    if (error) {
```
rhenshaw56: I don't see any case where this errorToast would be displayed since the action dispatched won't be doing any server side calls so we're sure of no server error for one.

app/js/components/orderEntry/OrdersTable.jsx
```diff
@@ -9,21 +9,86 @@ import DrugOrderDetails from './DrugOrderDetails';
...
+    const {
+      status: { added, error },
+      errorMessage,
+      labOrderData,
+    } = this.props.createLabOrderReducer;
     if (this.props.patient.uuid !== prevProps.patient.uuid) {
       this.props.dispatch(fetchOrders(null, this.props.patient.uuid));
     }
+    if (added && labOrderData !== prevProps.createLabOrderReducer.labOrderData) {
+      this.props.dispatch(toggleUrgency(
```
rhenshaw56: I don't think it makes sense to be dispatching an action inside of a component lifecycle, @IAMOTZ coman carry your brother

rhenshaw56: It's bug prone, especially if this implementation should be scaled in some way

app/js/components/orderEntry/OrdersTable.jsx
Efosaok: I guess you did not look at the action creator then. the toggling urgency implemented here is not just to update the draft data but the previously created order, so, an api call is definitely done in the action creator..

app/js/components/orderEntry/OrdersTable.jsx
rhenshaw56: well `added` doesn't really read so good, making it specific would make more sense

app/js/components/orderEntry/OrdersTable.jsx
rhenshaw56: this is the action for `toggleUrgency`
```
const toggleUrgency = (previousOrderId, newOrderId, newUrgency) => ({
  type: TOGGLE_URGENCY,
  previousOrderId,
  newOrderId,
  newUrgency,
});
```
so how is an api call definitely done here ? I guess I'm not understanding

app/js/components/orderEntry/OrdersTable.jsx
rhenshaw56: in your question `so, how and where is asynchronous API calls (fetching of data) done when a page is loaded done??` not when a component is been updating, preferably before initial render but definitely not when it's updating, and handler should be used to trigger updates. you can actually call `setState` in lifecycles which although it's still ill advised but definitely not in `render()`. Yeah it doesn't matter if it's not asynchronous or not, it's just not cool to update state state in the middle of an update, think about it...

Efosaok: your argument initially is that actions should not be dispatched in `componentLifeCycle` methods, the question I asked was just to bring to your attention that it is done and very necessary in a `React` project.

I also want to emphasize that I am not calling setState here, I am only dispatching an action that would trigger a `re-render` since the state would change. I only mentioned `setState` as an instance of the fact that a re-render can be triggered after a component is updated if necessary but should be done properly so that an update loop doesn't happen.

app/js/components/orderEntry/OrdersTable.jsx
Efosaok: did you look at the full code in this code block, `added` is a variable obtained from destructuring, It wasn't originally defined here.

and what is your idea of specific? please refer below to get an overview of what I am doing here.

Efosaok:
@rhenshaw56 

This is basically the workflow for this task.
first of all, the task here is toggling the urgency of an already created labOrder. it's almost thesame as editing a labOrder, so basically.

  - an API call is made to edit the lab Order, done in the `toggleUrgency` method.
  - the reducers act upon the action dispatched and manipulate the state, thus creating a component update.
  - if the urgency was successfully edited in the Resource, hence a success response received, instead of making an api call to fetch the whole data again cause the api call to edit returns only the edited order.an action is dispatched to edit the particular order in the store accordingly. and then a successfull toast is displayed, if an error response is received, an error toast is displayed as well, done in the `componentDidUpdate` method..

app/js/components/orderEntry/OrdersTable.jsx
Efosaok: please look at the toggleUrgency method, the last line is where the action that makes the api call is dispatched.

dkayiwa:
@Efosaok why is this still open, 12 days after?

dkayiwa:
Closing this until i get a response to the above.

PR #170, opened 2018-08-30
OEUI-238: Toggle lab urgency from shared draft

IAMOTZ:
### **JIRA TICKET NAME**
[OEUI-238: User should be able to toggle the urgency of a lab order from the draft.](https://issues.openmrs.org/browse/OEUI-238)

### **Summary**
This pull request adds the functionality that enables users to toggle the urgency of Lab Orders from the new shared draft.

### DIFF
--- a/app/js/actions/draftActions.js
+++ b/app/js/actions/draftActions.js
@@ -1,0 +1,26 @@
+import { TOGGLE_DRAFT_LAB_ORDER_URGENCY } from './actionTypes';
+import constants from '../utils/constants';
+
+const toggleDraftLabOrderUrgency = (order) => {
+  const urgencyTypeToggled = {
+    ROUTINE: constants.STAT,
+    STAT: constants.ROUTINE,
+  };
+  const orderUuid = order.uuid;
+  let orderUrgency;
+  const hasUrgencyProperty = Object.prototype.hasOwnProperty.call(order, 'urgency');
+  orderUrgency = constants.STAT;
+  if (hasUrgencyProperty) {
+    orderUrgency = urgencyTypeToggled[order.urgency];
+  }
+  const draftOrder = {
+    orderUuid,
+    orderUrgency,
+  };
+  return {
+    type: TOGGLE_DRAFT_LAB_ORDER_URGENCY,
+    order: draftOrder,
+  };
+};
+
+export default toggleDraftLabOrderUrgency;
--- a/app/js/components/orderEntry/OrderEntryPage.jsx
+++ b/app/js/components/orderEntry/OrderEntryPage.jsx
@@ -14,6 +14,7 @@ import getDateFormat from '../../actions/dateFormat';
 import activeOrderAction from '../../actions/activeOrderAction';
 import { fetchPatientRecord, fetchPatientNote } from '../../actions/patient';
 import { setSelectedOrder } from '../../actions/orderAction';
+import toggleDraftLabOrderUrgency from '../../actions/draftActions';
 import imageLoader from '../../../img/loading.gif';
 import './styles.scss';
 
@@ -59,6 +60,7 @@ export class OrderEntryPage extends PureComponent {
           handleDraftDiscard={() => {}}
           draftOrders={allDraftOrders}
           handleSubmit={() => {}}
+          toggleDraftLabOrderUrgency={this.props.toggleDraftLabOrderUrgency}
         />
       </div>);
   }
@@ -242,6 +244,7 @@ OrderEntryPage.propTypes = {
   currentOrderType: PropTypes.object,
   draftLabOrders: PropTypes.object.isRequired,
   draftDrugOrders: PropTypes.arrayOf(PropTypes.any).isRequired,
+  toggleDraftLabOrderUrgency: PropTypes.func.isRequired,
 };
 
 OrderEntryPage.defaultProps = {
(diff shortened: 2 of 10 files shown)


app/js/actions/draftActions.js
```diff
@@ -0,0 +1,25 @@
...
+  const orderUuid = order.uuid;
+  let orderUrgency;
+  const hasUrgencyProperty = Object.prototype.hasOwnProperty.call(order, 'urgency');
+  if (hasUrgencyProperty && order.urgency === constants.ROUTINE) {
+    orderUrgency = constants.STAT;
+  } else if (hasUrgencyProperty && order.urgency === constants.STAT) {
+    orderUrgency = constants.ROUTINE;
+  } else {
+    orderUrgency = constants.STAT;
+  }
```
Efosaok: since you are basically toggling between two values, I think the conditional blocks you are using is a bit of an overkill. please refactor to something like this.

```
let orderUrgency = constants.STAT
if (hasUrgency) {
   orderUrgency = constants[order.urgency]
}
```

app/js/components/orderEntry/OrderEntryPage.jsx
```diff
@@ -14,6 +14,7 @@ import getDateFormat from '../../actions/dateFormat';
 import activeOrderAction from '../../actions/activeOrderAction';
 import { fetchPatientRecord, fetchPatientNote } from '../../actions/patient';
 import { setSelectedOrder } from '../../actions/orderAction';
+import toggleDraftLabOrderUrgency from '../../actions/draftActions';
```
Efosaok: `name of import` should match `export-default`

tests/actions/draftActions.test.js
```diff
@@ -0,0 +1,57 @@
...
+        TOGGLE_DRAFT_LAB_ORDER_URGENCY,
+        order: {
+          ...labOrder, 
+          urgency: 'ROUTINE',
+        },
+      };
+      const store = mockStore({});
+      await store.dispatch(toggleDraftLabOrderUrgency(labOrder), () => {
+        expect(store.getActions()).toEqual(expectedAction);
+      });
```
Efosaok: `store.dispatch` is a `Promise` based function and to the best of my knowledge, doesn't take in a second parameter. since you are using the `await` keyword, refactor this to

```
await store.dispatch(toggleDraftLabOrderUrgency(labOrder));
expect(store.getActions()).toEqual(expectedAction);
 ```

File: tests/actions/draftActions.test.js [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
```
Efosaok: refer to above comment

tests/actions/draftActions.test.js
```diff
@@ -0,0 +1,57 @@
...
+        TOGGLE_DRAFT_LAB_ORDER_URGENCY,
+        order: {
+          ...labOrder, 
+          urgency: 'STAT',
+        },
+      };
+      const store = mockStore({});
+      await store.dispatch(toggleDraftLabOrderUrgency(labOrder), () => {
+        expect(store.getActions()).toEqual(expectedAction);
+      });
```
Efosaok: same here

app/js/actions/draftActions.js
IAMOTZ: Hi @Efosaok 
Yes, you are right, I am toggling between two values, however, the next value actually depends on the previous and that is why the series of checks is needed.
Let me try and explain the checks in plain English:
- If the order has urgency property and its previous value is ROUTINE, the urgency should become STAT.
- If the order has urgency property and its previous value is STAT, the urgency should become ROUTINE.
- If the order does not have urgency property, the urgency should become STAT.

I hope this clarifies your concern.

app/js/components/orderEntry/OrderEntryPage.jsx
IAMOTZ: The file `draftAction` is still going to contain a lot of actions that belong to the draft some of which are to be implemented in [this](https://issues.openmrs.org/browse/OEUI-236) and [this](https://issues.openmrs.org/browse/OEUI-239) tickets.
That is the reason why I named the file `draftActions`, once other actions are added to the file, the `export-default` won't be needed anymore.

app/js/actions/draftActions.js
Efosaok: I understand exactly what you are trying to do. but I think what I missed out is that you should set the `constant` variable to this.

```
export const STAT = 'ROUTINE'
export const ROUTINE = 'STAT'
```
and then just use the code up there to replace what you have already. this would make your  code more concise..

app/js/components/orderEntry/OrderEntryPage.jsx
Efosaok: If there was a rule to lint this, you know you'll probably have to fix it. I understand your point but the code written at any point in time should adhere to conventions used at that point in time.

app/js/actions/draftActions.js
IAMOTZ: Hmm..., now I get what you are saying.
I have updated the code with an implementation that reflects what you said above.
Thanks man, I was able to reduce some line of codes(and thinking effort) over there.

tests/actions/draftActions.test.js
IAMOTZ: Thanks.
The code has been updated!

app/js/components/orderEntry/OrderEntryPage.jsx
Efosaok: since this PR and the related PR's will be merges soonest, you can leave it as it is but just bear in mind that this is not acceptable under normal circumstances.

PR #171, opened 2018-08-30
OEUI-236: Edit drug orders from shared draft

IAMOTZ:
### **JIRA TICKET NAME**
[OEUI-236: User should be able to edit drug order from the draft](https://issues.openmrs.org/browse/OEUI-236)

### **Summary**
This pull request adds the functionality that enables users to edit drug orders from the new shared draft.


app/js/components/drugOrderEntry/index.jsx
```diff
@@ -35,6 +35,38 @@ export class SearchAndAddOrder extends React.PureComponent {
...
+        editDrugUuid: selectedOrder.drug,
+        editDrugName: selectedOrder.drugName,
+        orderNumber: selectedOrder.orderNumber,
+      });
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    const { selectedOrder, activity } = this.props;
+    if (activity === 'DAFT_ORDER_EDIT' && selectedOrder &&
```
Daramola98: Is this 'DAFT_ORDER_EDIT' a typo or is it on purpose?

IAMOTZ: Hi @Daramola98 
`DAFT_ORDER_EDIT` is actually a typo and surprisingly, I used it across all its location like that and that is why the functionality works 😄.
I would make sure to fix the typo.
Thanks!

Daramola98: Great

app/js/actions/orderAction.js
```diff
@@ -14,3 +14,7 @@ export const setSelectedOrder = orderDetails => ({
   type: SET_SELECTED_ORDER,
   ...orderDetails,
 });
+
+export const removeSelectedOrder = () => ({
```
rhenshaw56: you're dispatching `removeSelectedOrder` but I see just null values been set in state to null which doesn't really depicting a remove operation, don't you think it would be cool to just dispatch `SET_SELECTED_ORDER` which is already available and then pass the required arguments directly from the component ?

IAMOTZ: Yp @rhenshaw56 
Using the already existing `SET_SELECTED_ORDER` can actually help me achieve the same functionality. I never thought about using that approach.
Thanks for the feedback.

IAMOTZ: I have updated the code implementing you feedback @rhenshaw56.
Thanks again.

PR #173, opened 2018-08-30
OEUI-247: Implement Toast Message When Creating Drug Order

Efosaok:
### **JIRA TICKET NAME**
[OEUI-247](https://issues.openmrs.org/browse/OEUI-247)

### **Summary**
-  Use toast messages to display server messages when creating a drug order.

### DIFF
--- a/app/js/reducers/addDrugOrderReducer.js
+++ b/app/js/reducers/addDrugOrderReducer.js
@@ -5,7 +5,11 @@ import {
 
 const initialState = {
   addedOrder: {},
-  error: null,
+  errorMessage: '',
+  status: {
+    error: false,
+    added: false,
+  },
 };
 
 const addDrugOrderReducer = (state = initialState, action) => {
@@ -14,11 +18,20 @@ const addDrugOrderReducer = (state = initialState, action) => {
       return {
         ...state,
         addedOrder: action.data,
+        errorMessage: '',
+        status: {
+          error: false,
+          added: true,
+        },
       };
     case POST_DRUG_ORDER_FAILURE:
       return {
         ...state,
-        error: action.payload,
+        errorMessage: action.payload.response.data.error.message,
+        status: {
+          error: true,
+          added: false,
+        },
       };
     default:
       return state;
--- a/tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
+++ b/tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
@@ -2,7 +2,9 @@ import React from 'react';
 
 import { AddForm } from '../../../../app/js/components/drugOrderEntry/addForm/AddForm';
 
-const { draftOrders, draftOrder, editOrder, formType, session } = mockData;
+const {
+  draftOrders, draftOrder, editOrder, formType, session,
+} = mockData;
 
 const props = {
   addDraftOrder: jest.fn(),
@@ -15,13 +17,21 @@ const props = {
   setSelectedOrder: jest.fn(),
   getOrderEntryConfigurations: jest.fn(),
   allConfigurations: {
-    drugDosingUnits: [{display: 'grams'}],
-    orderFrequencies: [{display: 'daily'}],
-    drugRoutes: [{display: 'eye drops'}],
-    drugDispensingUnits: [{display: 'kits'}],
-    durationUnits:  [{display: 'months' }],
+    drugDosingUnits: [{ display: 'grams' }],
+    orderFrequencies: [{ display: 'daily' }],
+    drugRoutes: [{ display: 'eye drops' }],
+    drugDispensingUnits: [{ display: 'kits' }],
+    durationUnits: [{ display: 'months' }],
   },
-  careSetting: {bdisplay: 'Outpatient', uuid: 'aaa1234' },
+  addDrugOrderReducer: {
+    addedOrder: {},
+    errorMessage: '',
+    status: {
+      error: false,
+      added: false,
+    },
+  },
+  careSetting: { bdisplay: 'Outpatient', uuid: 'aaa1234' },
   drugName: 'Paracentamol',
   drugUuid: 'AJJJKW7378JHJ',
   draftOrder,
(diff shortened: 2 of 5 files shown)


Daramola98:
Is the JIRA Ticket name not meant to be the same as the one that is in PR header?

Efosaok:
@Daramola98  Yes, thanks for pointing that out..

app/js/reducers/addDrugOrderReducer.js
```diff
@@ -14,11 +18,19 @@ const addDrugOrderReducer = (state = initialState, action) => {
...
+        status: {
+          error: false,
+          added: true,
+        },
       };
     case POST_DRUG_ORDER_FAILURE:
       return {
         ...state,
-        error: action.payload,
+        errorMessage: action.payload.response.data.error.message,
```
Daramola98: maybe a liitle bit of destructing would be nice here to avoid long chains like this, what do you think @Efosaok

Efosaok: I am accessing this data in just one instance here, therefore, I see destructuring not really useful here.   😄

app/js/reducers/addDrugOrderReducer.js
```diff
@@ -14,11 +18,19 @@ const addDrugOrderReducer = (state = initialState, action) => {
       return {
         ...state,
         addedOrder: action.data,
+        status: {
```
IAMOTZ: Hi @Efosaok 
For `POST_DRUG_ORDER_SUCCESS`, are you not supposed to also set the `errorMessage` property to empty?

tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
```diff
@@ -232,5 +240,35 @@ describe('handleSubmitDrugForm() method', () => {
...
+      addDrugOrderReducer: {
+        errorMessage: '',
+        addedOrder: { id: 1, type: 'testorder' },
+        status: {
+          added: true,
+          error: false,
+        }
+      }
+    })
+    expect(global.toastrMessage).toEqual('Order Successfully Created')
```
IAMOTZ: Please assert that `global.toastrMessage` is not equal to `Order Successfully Created` before setting the props.

app/js/reducers/addDrugOrderReducer.js
Daramola98: Ok cool

app/js/reducers/addDrugOrderReducer.js
Efosaok: it's not necessary, I basically use the status object to check for error or success and you see that is what I update in both `action type` cases.

tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
Efosaok: It doesn't make sense that way as there is no reason to assert when no action has been made, assertions are made to ensure that a mocked action would always return the expected result, just asserting at the beginning of a test-case without a preceding action happening gives no context on what the assertion is for.

app/js/reducers/addDrugOrderReducer.js
IAMOTZ: Using the status object to check for error messages would definitely work fine, but I don't just think it would be nice to still have an error message in the state while the error status is `false` and a drug order has been successfully added.

tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
IAMOTZ: From the test description: `...displays a toast with a message if a drug has been added successfully...`, I believe this is what we want to test:
It is only when the drug is added(which you simulated with `wrapper.setPros({...})`) that the toast should appear with the message.

However, from the test implementation above, there is nothing that guarantees that `global.toastrMessage` does not alreay equals `Order Successfully Created` before you set the props. Adding an assertion to confirm that the `global.toastrMessage` does not contain `Order Successfully Created` before you set the props would give me enough confidence to believe that it is only after you set the props that the message appears.

IAMOTZ:
I just tested this functionality locally and it works fine 👍 
However, the error message that pops up does not seem to be explanatory enough to a User using the application. This is a snapshot: 

screen shot 2018-09-03 at 10 26 36 am

 I suggest you reformat the message to ensure that users fully understand the error toast.

app/js/reducers/addDrugOrderReducer.js
Efosaok: addressed

tests/components/DrugOrderEntry/addForm/AddForm.test.jsx
Efosaok: addressed

Efosaok:
@IAMOTZ 

you understand that the error message is coming from the API and the best I can do is remove the brackets around.

If the error messages from the API is not explicit, I am not in the position to reword or decide what the message should be..

topseySuave:
Please fix the conflicts

tests/reducers/addDrugOrderReducer.test.js
```diff
@@ -7,33 +7,47 @@ import addDrugOrderReducer from '../../app/js/reducers/addDrugOrderReducer';
...
         addedOrder: {},
-        error: null,
+        errorMessage: '',
+        status: {
+          error: false,
+          added: false,
+        },
       };
-    it('should return addedOrder payload', () => {
+    it(`parses payload from action to key addedOrder and sets status key
```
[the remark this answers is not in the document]
Efosaok: I used back-ticks because the string spans across multiple lines

tests/reducers/addDrugOrderReducer.test.js
```diff
@@ -7,33 +7,47 @@ import addDrugOrderReducer from '../../app/js/reducers/addDrugOrderReducer';
...
+              added: true,
+            }
         };
-        const newSate = addDrugOrderReducer(initialState, action);
-        expect(newSate).toEqual(expected);
+        const newState = addDrugOrderReducer(initialState, action);
+        expect(newState).toEqual(expectedState);
     });
-    it('should return an error', () => {
+    it(`parses payload error message from action to key errorMessage and
```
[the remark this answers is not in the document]
Efosaok: same reason stated above

IAMOTZ:
Nice one @Efosaok.
I actually did not think about it from that perspective.
Since this is an API thing, you can create a topic on Talk so the backend guys can attend to it, however, that shouldn't affect the merging of this Pull Request.

dkayiwa:
@Efosaok is this what you are looking for? https://github.com/openmrs/openmrs-core/blob/master/api/src/main/resources/messages.properties#L1533

Efosaok:
yes, That would help

dkayiwa:
I mean that is what you should use instead of that ugly error code.

Efosaok:
ok, Is there a query to use to get that actual user-defined error?

mogoodrich:
If we have a toast message pattern/library that we like, feel free to issue a pull request to openmrs-react-components for any functionality we can extract out.

We are using toastr in one of our PWAs, and I though I extracted it out into openmrs-react-components, but I can't currently find where I put it.

PR #187, opened 2018-09-13
OEUI-263: Add order from set to order types

Daramola98:
### **JIRA TICKET NAME**
[OEUI-263 Add order from sets to order types ](https://issues.openmrs.org/browse/OEUI-263)

### **Summary**
- Add order from sets to order types so that it displays in order type drop-down

### DIFF
--- a/app/js/components/orderEntry/RenderOrderType.jsx
+++ b/app/js/components/orderEntry/RenderOrderType.jsx
@@ -19,6 +19,9 @@ const RenderOrderType = (props) => {
         />
       );
     }
+    case orderTypes.ORDER_FROM_SETS.id: {
+      return "<InsertComponent />";
+    }
     default: {
       return (<AllOrders />);
     }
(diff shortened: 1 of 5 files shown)


app/js/components/orderEntry/RenderOrderType.jsx
```diff
@@ -19,6 +19,9 @@ const RenderOrderType = (props) => {
         />
       );
     }
+    case orderTypes.ORDER_FROM_SETS.id: {
+      return "<InsertComponent />";
```
topseySuave: WHy is the name of the component `InsertComponent`? Is there any particular reason?

Daramola98: It is just a placeholder for now, when the new component is created it would be replaced

topseySuave: Or do you just mean to place it there after the component is ready?

topseySuave: Ok, great

app/js/components/orderEntry/SortAndFilter.jsx
```diff
@@ -14,6 +14,7 @@ export const SortAndFilter = props => (
         <option value="all">All</option>
         <option value="drugorder">Drug Orders</option>
         <option value="testorder">Test Orders</option>
+        <option value="orderset">Order from Sets</option>
```
IAMOTZ: `Order from Sets` is not a new type of order, it is just a way of grouping orders together. So I don't think there should be a functionality like sorting with "Order from Sets".

Daramola98: Ok will remove it

PR #189, opened 2018-09-14
OEUI-271: Remove Pre existing save drug order functionality

Rhotimee:
### **JIRA TICKET NAME**
[OEUI-271: Remove Pre-existing save drug order functionality](https://issues.openmrs.org/browse/OEUI-271)

### **Summary**

- The purpose of this task is to remove the redundant code from the codebase.
addDrugOrder and createLabOrder code was removed from the codebase (alongside thier actions and reducers) and createOrder was used instead. As the orders are no longer separate but consolidated.


app/js/actions/createOrder.js
```diff
@@ -3,9 +3,9 @@ import axiosInstance from '../config';
 
 import { SAVE_DRAFT_LAB_ORDER } from './actionTypes';
 
-const createLabOrder = labOrderData => ({
+const createOrder = labOrderData => ({
```
Efosaok: the parameter name here should also change to a more befitting variable name.

Rhotimee:
The test coverage reduced because of the test files that were removed.

app/js/actions/createOrder.js
Rhotimee: Resolved.

PR #193, opened 2018-09-26
OEUI-282: (New) and (Revise) should not appear on order list

Daramola98:
### **JIRA TICKET NAME**
[OEUI-282: (New) and (Revise) should not appear on order list](https://issues.openmrs.org/browse/OEUI-282)

### **Summary**
- Remove (New) and (Revise) from the order list
- New and Revise should show on draft list

### DIFF
--- a/app/js/components/orderEntry/OrdersTable.jsx
+++ b/app/js/components/orderEntry/OrdersTable.jsx
@@ -163,35 +163,38 @@ export class OrdersTable extends PureComponent {
       <React.Fragment>
         {filteredOrders &&
           filteredOrders.length !== 0 &&
-          filteredOrders.map(order => (
-            <Accordion
-              title={
-                <OrderHeader
-                  status="Active"
-                  orderable={order.display}
-                  order={order}
-                  handleEdit={this.handleActiveOrderEdit}
-                  handleDiscontinue={this.setDiscontinuedDrugOrder}
-                />
-              }
-              key={order.uuid}
-              dateFormat={dateFormat}
-              date={order.dateActivated}>
-              {order.type === 'drugorder' ? (
-                <DrugOrderDetails
-                  dosingInstructions={order.dosingInstructions}
-                  dispense={order.dispense}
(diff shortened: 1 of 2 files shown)


dkayiwa:
Any tests to accompany this change?

Daramola98:
> Any tests to accompany this change?

No tests has it is data coming from the API

dkayiwa:
I do not understand your response.

Daramola98:
No functionality changed just manipulating data that is displayed, would be like testing if the string methods used does what they are supposed to do

dkayiwa:
Do you mean we cannot have tests to ensure that "(New) and (Revise) should not appear on order list"?

Daramola98:
So the data that comes from the API contains (NEW) or (REVISE), that data is passed to the component and it's been manipulated in the component to remove them, we can have test that ensure this but snapshots were used for testing components but I can write a test case that would handle that, but it would never fail because it would be mock data that would be passed to the props, so its just like testing if the string I am passing to props which contain (NEW) or (REVISE) responds to the substring method.

Daramola98:
what do you think? should I go ahead an write a test that does this?

Daramola98:
Ok will do this, Thanks

dkayiwa:
But does it make sense to you?

Daramola98:
Yes it does, will have to make sure the value of that column does not have NEW or REVISE

mseaton:
@Daramola98 and @dkayiwa just noticing this conversation and a bit concerned that we are parsing a display property of a REST response in order to get the information we need here.  Can't we get the actual orderable and it's display property properly, rather than resorting to a somewhat hacky need to substring?

Daramola98:
I just checked the rest response again and there seems to be a drug display object that contains just the drug name, I think I would refactor to use just that instead of the substring method @dkayiwa @mseaton

mseaton:
@Daramola98 perfect.

Daramola98:
@mseaton @dkayiwa I have made the necessary changes, Please review Thanks

PR #194, opened 2018-09-27
OEUI-280: UI Improvements on Order List page

Rhotimee:
### **JIRA TICKET NAME**
[OEUI-280 UI Improvements on Order List page](https://issues.openmrs.org/browse/OEUI-280)

### **Summary**
- The order name was left-justified in the details column
- Only the date, not the time, is showed in the date column.  This eliminated wrapping and allow for the rows to be thinner.  
- The entire date and time can show when the row is expanded.
- For drugs, only the drug name is showing.
- The dosage, frequency etc....  are all show when the row is expanded.

### DIFF
--- a/app/js/components/orderEntry/Accordion.jsx
+++ b/app/js/components/orderEntry/Accordion.jsx
@@ -26,7 +26,7 @@ class Accordion extends React.Component {
                 <i className="toggle-icon icon-caret-right small rotate90" />
               )}
             </span>
-            {format(date, dateFormat)}
+            {format(date, dateFormat).slice(0, -5)}
           </th>
           {this.props.title}
         </tr>
--- a/app/js/components/orderEntry/LabOrderDetails.jsx
+++ b/app/js/components/orderEntry/LabOrderDetails.jsx
@@ -1,11 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const LabOrderDetails = ({ urgency, orderer }) => (
+const LabOrderDetails = ({ urgency, orderer, activeDates }) => (
   <div className="order-details">
     <div className="details">
       <ul>
         <b>Urgency:</b> {urgency}
+      </ul>
+      <br />
+      <ul>
+        <b>Active Dates:</b> {activeDates}
       </ul>
       <br />
       Ordered by {orderer}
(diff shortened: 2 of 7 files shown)


tests/components/orderentry/OrdersTable.test.jsx
```diff
@@ -218,4 +218,4 @@ describe('discontinueDrugOrder method', () => {
     expect(props.dispatch).toBeCalled();
     done();
   });
-});
+});
```
topseySuave: EOL error here @Rhotimee

app/js/components/orderEntry/Accordion.jsx
```diff
@@ -26,7 +26,7 @@ class Accordion extends React.Component {
                 <i className="toggle-icon icon-caret-right small rotate90" />
               )}
             </span>
-            {format(date, dateFormat)}
+            {format(date, dateFormat).slice(0, -5)}
```
Efosaok: hey, why slicing through the dates, the `date format`is a global property that is set in every standalone. so this might look well on your standalone but not good in another's. so I'd suggest you remove the slicing and come up with a general dateformat every user should set their standalone to.

app/js/components/orderEntry/LabOrderDetails.jsx
```diff
@@ -1,13 +1,17 @@
...
-const LabOrderDetails = ({ urgency, orderer }) => (
+const LabOrderDetails = ({ urgency, orderer, activeDates }) => (
   <div className="order-details">
     <div className="details">
       <ul>
         <b>Urgency:</b> {urgency}
       </ul>
       <br />
+      <ul>
+        <b>Active Dates:</b> {activeDates}
```
Efosaok: is this present in the mockup ?

app/js/components/orderEntry/Accordion.jsx
Rhotimee: The standard date format is DD-MM-YYYY HH:mm. Therefore, I removed the Time in this case.

app/js/components/orderEntry/LabOrderDetails.jsx
Rhotimee: David requested for it to be present there. See here https://issues.openmrs.org/browse/OEUI-280

PR #207, opened 2018-10-25
OEUI-293: Enforce prop-type validation in all components

Daramola98:
### **JIRA TICKET NAME**
[OEUI-293: Enforce prop-type validation in all components](https://issues.openmrs.org/browse/OEUI-293)

### **Summary**
- Some components do not have prop-type validation present, the goal of this ticket is to locate such components and add prop-type validations.


Tittoh:
LGTM

dkayiwa:
Are these the only components that need prop-type validation?

Daramola98:
> Are these the only components that need prop-type validation?

All other components already have prop-type validation, these components did not have prop-type validation or some props were missing so I added them.

PR #216, opened 2018-12-12
fix failing bamboo build

rhenshaw56:
WOR-205: orderentry-owa build failling


mogoodrich:
@rhenshaw56 defer to you on what version of uglify to use, I don't have any experience with it!

PR #221, opened 2019-04-18
OEUI-305: prevent creating multiple encounters when cancelling orders

rhenshaw56:
### **JIRA TICKET NAME**
[OEUI-305](https://issues.openmrs.org/browse/OEUI-305) Cancelling an Order should use existing encounter if present

### **Summary**
- sample summary


app/js/rest/encounterRest.js
```diff
@@ -0,0 +1,35 @@
+
+import axiosInstance from '../config';
+
+const api = {
```
mogoodrich: Ideally, we'd use the encounter REST from react-components, but you were saying there was a problem with the format that the encounterRest in react-components returned? Or was it something else?

app/js/rest/encounterRest.js
```diff
@@ -0,0 +1,35 @@
...
+  fetchEncountersByPatient: (patient, encounterType, provider, location) => axiosInstance.get(`encounter/?patient=${patient}&encounterType=${encounterType}&provider=${provider}&location=${location}&v=full`)
```
mogoodrich: "Full" represetations are pretty huge... if we can fetch a custom rep that meets our needs closely, that would be better. Don't we really just need the encounterDatetime of the encounters?

app/js/sagas/orderSaga.js
```diff
@@ -0,0 +1,82 @@
...
+    if (matchingEncounter.matched) {
+      encounterPayload = {
+        ...encounterPayload,
+        uuid: matchingEncounter.encounter.uuid,
+      };
+      yield call(encounterRest.updateEncounter, encounterPayload);
+    } else {
+      yield call(encounterRest.createEncounter, encounterPayload);
+    }
+    window.location.reload();
```
mogoodrich: Maybe there's no way around it, but it seems weird to be triggering a window reload within the saga?

app/js/sagas/orderSaga.js
```diff
@@ -0,0 +1,82 @@
...
+  const provider = state.encounterRoleReducer.encounterRole.uuid;
+  const duration = format(addMinutes(new Date(), -60));
+
+  try {
+    const encounterResponse = yield call(
+      encounterRest.fetchEncountersByPatient,
+      patientUuid,
+      encounterTypeUuid,
+      provider,
+      locationUuid,
```
mogoodrich: It would be better if we could pass in the date here as well, to restrict the size of the returned respond, but if we currently don't have an endpoint that supports dates, I'd be fine with this for now.  The main thing would be to see if we can reduce it not to include the full representation of the encounter... for a long-time patient, they might have 50+ encounters with the same provider and this would be a huge response payload.

app/js/sagas/orderSaga.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+    if (encounterResponse.results.length) {
+      const matched = getDateRange(
+        encounterResponse.results,
+        duration,
+        new Date(),
+        'encounterDatetime',
+      );
+
+      if (matched.length) {
```
mogoodrich: I'd add a comment here like "if we find multiple matching encounters, just return first one"  (which I think is fine, but good to note that we realize we are just taking an arbitrary encounter in case we want to refine this logic in the future.

app/js/sagas/orderSaga.js
rhenshaw56: @mogoodrich tried that but it doesn't seem to work, it still returns all the results. maybe i'm not doing it right

app/js/rest/encounterRest.js
rhenshaw56: @mogoodrich yeah, it was returning the location as null

mogoodrich: Strange, it looks like the default REST rep for Encounter in react-components includes "location"... (and it should).

app/js/sagas/orderSaga.js
mogoodrich: I can take a look...

PR #226, opened 2019-09-05
OEUI-307: Prevent cancellation of Test Orders that are Pending or Com…

mogoodrich:
…pleted


cioan:
Looks good @mogoodrich ! Thanks.

PR #228, opened 2019-09-27
UHM-4018

charlesdupuyrony:
### **JIRA TICKET NAME**
[UHM-4018](https://mocklink) mock text

### **Summary**
- Added column order number in the patient orders ui

- [ ] I can sucessfully create Drug orders
- [ ] I can sucessfully create Lab orders
- [ ] I can sucessfully search for drug orders
- [ ] I can sucessfully check all tests passed

### DIFF
--- a/app/js/components/orderEntry/SortAndFilter.jsx
+++ b/app/js/components/orderEntry/SortAndFilter.jsx
@@ -16,25 +16,15 @@ export const SortAndFilter = props => (
       <select
         className="sort-select"
         onChange={event => props.sortAndFilterAction('type', event.target.value)}>
-        Type:
-        <option value="all">
-          <FormattedMessage
-            id="app.orders.all"
-            defaultMessage="All"
-            description="All" />
-        </option>
-        <option value="drugorder">
-          <FormattedMessage
-            id="app.orders.drug"
-            defaultMessage="Drug Orders"
-            description="Drug Orders" />
-        </option>
-        <option value="testorder">
-          <FormattedMessage
-            id="app.orders.test"
-            defaultMessage="Test Orders"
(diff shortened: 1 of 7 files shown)


app/js/components/orderEntry/SortAndFilter.jsx
```diff
@@ -16,25 +16,15 @@ export const SortAndFilter = props => (
...
-        </option>
+        <FormattedMessage id="app.orders.all" defaultMessage="All" description="All">
+          {m => <option value="all">{m}</option>}
+        </FormattedMessage>
+        <FormattedMessage id="app.orders.drug" defaultMessage="Drug Orders" description="Drug Orders">
+          {m => <option value="drugorder">{m}</option>}
+        </FormattedMessage>
+        <FormattedMessage id="app.orders.test" defaultMessage="Test Orders" description="Test Orders">
+          {m => <option value="testorder">{m}</option>}
+        </FormattedMessage>
```
mogoodrich: This is great, I missed that FormatMessage supports a function as a child... I think there are other use cases where this will be very helpful.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "orderentry",
```
mogoodrich: I'm wondering if one of these chnanges (which assumedly happened after a npm install) were the source of the failing testing... one thing I don't think the rest of us tried was running an npm  install and then running the test... will be interesting to see if that test fails on Bamboo after this merges... if it does, I think it would likely mean one of two things:

* There was a bug in one of the dependency upgrades that introduced an error, or
* There was a change in one of the dependencies that exposed a bug in our code or our tests...

PR #229, opened 2019-09-30
UHM-4073: No confirmation displayed when order is canceled

mogoodrich:


cioan:
Looks good @mogoodrich ! Can you look at the travis builds to see if there is anything we need to worry? Thanks.

mogoodrich:
I cut a new version of React Components, and this fixed the Travis failure... thanks @cioan , I am merging now...

PR #230, opened 2019-10-03
Uhm 4018

charlesdupuyrony:
### **JIRA TICKET NAME**
[OEUI-1234](https://mocklink) mock text

### **Summary**
- sample summary

## I have checked that on this Pull request

- [ ] I can sucessfully create Drug orders
- [ ] I can sucessfully create Lab orders
- [ ] I can sucessfully search for drug orders

PR #231, opened 2019-10-11
UHM-4018

charlesdupuyrony:
### **JIRA TICKET NAME**
[OEUI-1234](https://mocklink) mock text

### **Summary**
- sample summary

- [ ] I can sucessfully create Drug orders
- [ ] I can sucessfully search for drug orders


app/js/components/orderEntry/OrderEntryPage.jsx
```diff
@@ -300,34 +315,25 @@ export class OrderEntryPage extends PureComponent {
...
-            </div>
+            <Tabs
+              selectedIndex={this.state.index} onSelect={ (i) => {
+                this.switchTabs(i); this.setState({ index: i });
+              }
+              }>
+              <TabList>
+                <Tab><FormattedMessage id="app.orders.list" defaultMessage="Orders List" description="Orders List" /></Tab>
+                <Tab><FormattedMessage id="app.orders.labs.add" defaultMessage="Add Lab Orders" description="Add Lab Orders" /></Tab>
+              </TabList>
```
mogoodrich: @charlesdupuyrony I still think we are missing the Tab here for Drug Orders, see my comment here:

https://github.com/openmrs/openmrs-owa-orderentry/pull/230/files#r331563422

Let me know if I'm just missing it, and I can test further, thanks!

I believe you can see the SelectOrderType and RenderOrderType files for how this currently works.

mogoodrich:
@charlesdupuyrony so I ran the latest code and the "Drug Orders" tab is showing up, which is what should happen by default, but unfortunately it's showing up even if "page=laborders" is passed in... in that case, only the "Lab orders" tab should show (and vice versa... if page=drugorders only the drug tab should show)



Also, this is strange, but there are some alignment issues I'm seeing today that I wasn't seeing yesterday... the "order list" is squeezed and there's no margin/padding for some of the items.  Was there a CSS change?  Thanks!

PR #235, opened 2020-05-15
Test snapshots should be committed and checked

brandones:


app/js/components/Draft.jsx
```diff
@@ -34,7 +33,7 @@ export class Draft extends PureComponent {
       );
 
       return (
-        <li className="draft-list small-font" key={shortid.generate()}>
```
mogoodrich, line 37: Ah, this to make the snapshots consistent?

brandones: Yep!

PR #241, opened 2020-06-17
Remove extra return button; add edit and delete buttons to drug orders

brandones:
https://talk.openmrs.org/t/order-entry-ui-return-button-not-working
https://issues.openmrs.org/browse/OEUI-308

Screenshot from 2020-06-17 14-57-28

The behavior of the cancel button is very irregular. Sometimes it discontinues the order. Sometimes it causes the server to 500 with `org.openmrs.api.OrderEntryException: Order.type.does.not.match`. Sometimes it causes a 500 with something like "Cannot discontinue an already discontinued order." Presumably these are client errors and should 400, but that's tangential.

The bugginess probably stems from the structure of the Redux store. It has one store per reducer, which means that there is no canonical data source anywhere. It is guaranteed by design to produce inconsistency. I don't think solving those problems is within the scope of this ticket.


mogoodrich:
Thanks @brandones !

Yes, unfortunately the bigger issues with cancelling drug orders seems out of scope of this ticket.  I wanted focus on the issues that we (PIH) introduced, which included the double return button, and also do the quick fix where it wasn't rendering the "Cancel" button for drug orders because it wasn't parsing the order type properly when the order was  drug.

If there are bigger issues with cancelling drug orders, I think this is an existing bug that has existing for some time (@Ruhanga do you know if this ever worked?) and we should prioritize fixing it but it's a separate scope of work.

We may want to decide, however, just *not* to allow cancelling drug orders until we fix the issue.  We at PIH aren't currently using drug ordering so I defer to people who are.

Ruhanga:
Thanks @brandones, @mogoodrich. We may then cover concealing/editing of an order in a different ticket.

brandones:
Sounds good @Ruhanga . Like I suggested above, it will probably be worth it just to either
1. totally rearrange the Redux state so that there actually is a single, unified state and not a state-per-reducer or
2. totally replace Redux with native React state management

I probably won't have time to do that in the very near future, and it would probably be a little more up-front investment, but I think it will probably be a better path than forever playing whack-a-mole with bugs in the present architecture. I could be available for some pairing and code review if there is someone available to take that on. (@mogoodrich and @mseaton might have other ideas though).

mogoodrich:
Although we (PIH) aren't currently using the Drug Order functionality, that is definitely something we'd like to use in the future (and hopefully the not-so-distance future!).   @brandones one interesting thing to consider would be whether this is likely to be superceded by something in the SPA application, and/or if we'd want to migrate this into a SPA app while making these fixes.

@Ruhanga just curious, where are you currently using this app?

brandones:
Yeah I think migrating to a SPA app would be great, assuming this module continues to be a good fit for our business requirements.

Ruhanga:
Thanks @mogoodrich, @brandones. We do not currently use the order entry app in our distributions. I'm simply following up on this because of the growing interest in making drug/lab orders by some community members. I can see this on the next Ref app [roadmap](https://wiki.openmrs.org/display/docs/Technical+Roadmap#TechnicalRoadmap-ReferenceApplication2.11(TBD)) as well.

PR #243, opened 2020-10-21
UHM-4934: Create "circular workflow" for adding Lab Orders from the L…

mogoodrich:
https://pihemr.atlassian.net/browse/UHM-4934

### DIFF
--- a/app/js/components/orderEntry/OrderEntryPage.jsx
+++ b/app/js/components/orderEntry/OrderEntryPage.jsx
@@ -65,9 +65,10 @@ export class OrderEntryPage extends PureComponent {
     if (added && labOrderData !== prevProps.createOrderReducer.labOrderData) {
       successToast(orderCreatedMsg);
       // if an afterAddOrderUrl has been specified, redirect to that page
-      if (afterAddOrderUrl && labOrderData && labOrderData.orders && labOrderData.orders.length > 0) {
-        // we pass in the uuid of the new order (just picking the "first" order if multiple were created)
-        let url = Handlebars.compile(afterAddOrderUrl)({
+      if (afterAddOrderUrl && labOrderData &&
+          labOrderData.orders && labOrderData.orders.length > 0) {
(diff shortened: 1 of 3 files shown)


app/js/components/orderEntry/OrderEntryPage.jsx
```diff
@@ -58,10 +60,20 @@ export class OrderEntryPage extends PureComponent {
...
+    const { afterAddOrderUrl } = this.state;
     const orderCreatedMsg = intl.formatMessage({ id: "app.orders.create.success", defaultMessage: "Order Successfully Created" });
     if (added && labOrderData !== prevProps.createOrderReducer.labOrderData) {
       successToast(orderCreatedMsg);
-      this.props.fetchLabOrders(null, this.props.patient.uuid);
+      // if an afterAddOrderUrl has been specified, redirect to that page
+      if (afterAddOrderUrl && labOrderData && labOrderData.orders) {
+        // we pass in the uuid of the new order (just picking the "first" order if multiple were created)
+        let url = Handlebars.compile(afterAddOrderUrl)({
+          order: labOrderData.orders[0].uuid,
```
brandones, line 71: Ok as long as you're confident that `labOrderData.orders` cannot be an empty array, and that if it has a member, that member must have a `uuid` element.

brandones (approved):

brandones:
Oh yeah, gotta fix the build too. Looks like you got a couple lint failures. Use `npm run lint` locally and fix stuff.

app/js/components/orderEntry/OrderEntryPage.jsx
mogoodrich: I think so, but, fair enough, I'll add a check.

mogoodrich: Actually, it really *must* have a uuid, but good to check that is has a member.

mogoodrich:
Fixed the lint issues... merging, thanks @brandones !

PR #247, opened 2021-04-02
GitHub actions

mseaton:
Initial attempt to do the following:

* Remove dependence on Travis
* Remove dependence on Bintray

This migrates to github actions to:
* Build, test, and attempt to do coveralls on each push and PR, deploy a snapshot of zip to maven repo
* Build, test, and deploy a non-snapshot zip on each tag to maven repo

The non-snapshot releasing has not yet been tested, but see the Actions tab to see the recent success with other aspects.

Some things still to do / figure out:
* Can we identify github releases differently from tags?  Should we use a naming convention to identify (vx.y.z?)
* Need to replace the maven credentials and repositories with non-PIH versions

@ibacher and @mogoodrich and @bmamlin FYI, feedback appreciated.


ibacher (approved):
Some minor comments, but this looks like a decent approach to me.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,48 @@
...
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [ 13.x ]
```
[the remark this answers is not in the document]
mseaton: Fine with me, as long as the build passes.  I couldn't find any documentation as to what version of Node was required to run this, and I always find that problematic.  I ended up going to our PIH Bamboo server that was building this successfully and it was using 13.11.0, so I went with that.  Will try updating to 14 and see what happens.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - run: npm ci
+      - run: npm run build --if-present
+      - run: npm test
+
+      - name: Coveralls
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
```
mseaton, line 35: FYI @ibacher - I don't actually know if this is enough to do the Coveralls correctly.   I was following https://github.com/marketplace/actions/coveralls-github-action and https://github.com/nickmerwin/node-coveralls and it was unclear to me whether I needed the "finish" step or not.  When I tried adding the finish step in it failed, so I removed it, but not sure that means that things are not actually finishing.  I don't really think Coveralls support should be a dealbreaker anyway - I'm fine getting rid of it personally, especially since this entire repo is deprecated.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,48 @@
...
+          java-version: 1.8
+          server-id: openmrs-repo-modules
+          server-username: MAVEN_REPO_USERNAME
+          server-password: MAVEN_REPO_PASSWORD
+
+      - name: Deploy release
+        run: mvn deploy:deploy-file -Pdeploy-snapshot
+        env:
+          MAVEN_REPO_USERNAME: pih
+          MAVEN_REPO_PASSWORD: ${{ secrets.MAVEN_REPO_PASSWORD }}
```
mseaton, line 48: @ibacher / @cintiadr / @bmamlin - I'll want to have a MAVEN_REPO_PASSWORD and MAVEN_REPO_USERNAME that we can use for deployment, ideally as organization-wide secrets that are shared with all of our OWA projects.  Can we do this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openmrs/orderentry-owa",
-  "version": "1.2.3",
+  "version": "1.2.4-SNAPSHOT",
   "description": "OWA for placing an order, revising it, discontinuing it, and being able to look up a patient&#39;s active orders.",
```
mseaton, line 4: No idea if the 'SNAPSHOT' convention is frowned upon in the NPM world, but it is consistent with our Maven packaging and would seem to fit well with the idea of hosting in Maven Repo

pom.xml
```diff
@@ -0,0 +1,101 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.openmrs.owa</groupId>
+    <artifactId>orderentry</artifactId>
+    <version>1.2.4-SNAPSHOT</version>
```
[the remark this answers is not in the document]
mseaton: @ibacher - I agree, and that plugin is a cool find, I'd definitely be up for using it.  I was also worried about this at first, but then realized that:
a) We were already maintaining this version twice between package.json and bintray.json
b) The deploy itself will fail if these don't match, as the deploy step is looking for a file named orderentry-<maven-version>.jar, which will only succeed if these are aligned.

So I think b) covers us in the sense that it will be obvious when running a build or trying to deploy that these don't match and the build will fail.

pom.xml
```diff
@@ -0,0 +1,101 @@
...
+                <enabled>false</enabled>
+            </snapshots>
+        </pluginRepository>
+    </pluginRepositories>
+
+    <distributionManagement>
+        <repository>
+            <id>openmrs-repo-modules</id>
+            <name>OpenMRS Modules</name>
+            <url>https://openmrs.jfrog.io/artifactory/modules-pih/</url>
```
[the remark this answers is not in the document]
mseaton: Is this a new thing you added?

mseaton (commented):
@ibacher thanks for the review - some reply comments...

.github/workflows/build-and-test.yml
mseaton: Follow-up, since I have access to the main openmrs repos, I configured this with my personal credentials and an API key for my account temporarily.  This seems to work, and on my last push to this feature branch this successfully deployed to the owa-snapshots repository.  Thanks for setting that up @ibacher .

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
mseaton: I've updated to version 14.x, and everything still builds fine.  Thanks @ibacher for the advice.  Will resolve this conversations.

pom.xml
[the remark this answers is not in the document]
mseaton: I've changed to use the owa repo as my comment above indicates.  Thanks for adding this!

pom.xml
```diff
@@ -0,0 +1,101 @@
...
+        </repository>
+        <snapshotRepository>
+            <id>openmrs-repo-modules</id>
+            <name>OpenMRS PIH Module Snapshots</name>
+            <url>https://openmrs.jfrog.io/artifactory/modules-pih-snapshots/</url>
```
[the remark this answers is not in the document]
mseaton: I've updated this as suggested.

package.json
mogoodrich: @ibacher  @brandones  @bmamlin  thoughts on this?

brandones: I don't believe that versioning scheme will work with npm (`npm version` & `npm publish`) but probably that doesn't matter if you're not publishing using npm.

ibacher: Just to reinforce what @brandones said, `1.2.4-SNAPSHOT` is technically a valid SemVer, but in Maven land `-SNAPSHOT` functions more like a tag than a version identifier (so `1.2.4-SNAPSHOT` might resolve to `1.2.4-20210402.193656-5` or `1.2.4-20210402.191113-4` depending on when it's run). `npm` just doesn't use that convention, so it would want to distinguish between `1.2.4-SNAPSHOT.0` and `1.2.4-SNAPSHOT.1`. But assuming we're just using this internally and publishing to Maven (which will handle the `-SNAPSHOT` part), it's probably not that big of a deal.

mseaton: If there is a better place for us to publish both the latest build and all of our releases for a given owa, and for the latest builds to be available through a consistent tag/alias that doesn't require updating our distribution to depend on a new build number of every commit, I think we should consider it.  Given that this is exactly what Maven does, and we use Maven successfully this way with existing tooling for our omod and war artifacts, this seems like the best way to do this to me, at least for now.
We will need our SDK to be updated to support pulling OWAs out of Maven rather than (or in addition to) Bintray.

ibacher: I'm totally 100% in support of just using `-SNAPSHOT` in the version name for OWAs.

